PR #23, opened 2023-11-26
Feat(web): Add ESlint rule to prevent use of 'process.env' within the source code.

sharma-shanu:
Since we are using 't3-env' to support typesafe environment variables across the web app.

This PR adds an ESlint rule in the apps/web project to avoid the direct access of environment variables in the source code.

During Development: 
Screenshot 2023-11-27 at 5 09 09 AM

After  `npm run lint`:

Screenshot 2023-11-27 at 5 10 21 AM


elie222:
Nice 😄

PR #36, opened 2023-12-19
Easier local setup

atbe:
- Include a `docker-compose` file which can standup a local redis & postgres server
- Include a link to the Tinybird README, with a mention to it in the main README
- Updated the `.env.example` to include the required `DIRECT_URL` env var -- its required for the prisma migration tool
- Default the redis host env var to local redis
- Don't require redis token, given that you can use local redis which has no auth by default


elie222:
Thanks!

zxt-tzx:
> - `TINYBIRD_TOKEN` -- Admin token for your Tinybird workspace (be sure to create an instance in the GCP `us-east4` region).

Just checking on this: is the region part absolutely necessary?

PR #42, opened 2023-12-24
Remove redundant port exposes from docker-compose.yml

sharma-shanu:
This PR :-

- Removes redundant port exposes for redis and postgres
- Adds service name to avoid naming clashes for docker networks and containers.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,8 @@
 version: "3.5"
+name: inbox-zero-services
 services:
   db:
-    image: "postgres"
+    image: postgres
     restart: always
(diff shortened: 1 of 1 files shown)


sharma-shanu:
The port exposed by docker-compose are already being exposed by the base image.

Reference :-

https://github.com/docker-library/postgres/blob/c86568af4a6861cb30b8f1b736b0868a3129bdd6/16/bookworm/Dockerfile#L221

https://github.com/docker-library/redis/blob/76c393fd8ac537201fe4bc4f200b4b5daee03f24/7.2/debian/Dockerfile#L151

Issue #81, opened 2024-01-05
Mark emails as read


elie222:
- [x] Mark an email when it's read
- [ ] Ability to mark an email as unread

elie222:
`UNREAD` label:
https://developers.google.com/gmail/api/guides/labels
https://developers.google.com/gmail/api/reference/rest/v1/users.messages/modify

See labelling in:
`apps/web/utils/gmail/label.ts`
Can add a function to this file: `markRead(read: boolean)`

In Gmail, messages receive labels. When a thread receives a label it's like applying it to every message in the thread.

ThisIsRahmat:
Hi @elie222 can I work on this?

elie222:
> Hi @elie222 can I work on this?

Hey, I missed this. It's been half done now. The other part is marking as unread. Although probably better to take a different issue. I'll probably add this myself soon.

PR #83, opened 2024-01-07
#fix:81 Mark emails as read

Shalom2212:
Created a mail read and unread function
added a button in ActionButtons where a user can Read or Unread a mail
contribute:issue #81 (1)
contribute:issue #81 (2)


apps/web/components/ActionButtons.tsx
```diff
@@ -56,6 +61,20 @@ export function ActionButtons(props: {
     setIsTrashing(false);
   }, [threadId, refetch]);
 
+  const [isReadUnreading, setisReadUnreading] = useState(false);
```
elie222: `setisReadUnreading` => `setIs`

apps/web/components/ActionButtons.tsx
```diff
@@ -56,6 +61,20 @@ export function ActionButtons(props: {
...
+  const [isStateRead, setIsStateRead] = useState(isRead);
+  useMemo(() => {
+    setIsStateRead(isRead);
+  }, [isRead]);
```
elie222: Not sure what the goal is here but this isn't how `useMemo` is used.

Learn more here: https://react.dev/reference/react/useMemo

apps/web/components/ActionButtons.tsx
```diff
@@ -56,6 +61,20 @@ export function ActionButtons(props: {
     setIsTrashing(false);
   }, [threadId, refetch]);
 
+  const [isReadUnreading, setisReadUnreading] = useState(false);
+  const [isStateRead, setIsStateRead] = useState(isRead);
```
elie222: Don't think you want to use this local state at all. You want to know the state of the email from the row itself. We use it other places to mark the row color as read.

apps/web/utils/gmail/label.ts
```diff
@@ -15,3 +15,33 @@ export async function labelThread(options: {
...
+  const { gmail, threadId } = options;
+
+  return gmail.users.threads.modify({
+    userId: "me",
+    id: threadId,
+    requestBody: {
+      removeLabelIds: ["UNREAD"],
+    },
+  });
+}
```
elie222: I'd be okay with these two functions being one. But not critical.

apps/web/utils/gmail/label.ts
```diff
@@ -15,3 +15,33 @@ export async function labelThread(options: {
     },
   });
 }
+
+export async function markReadLabel(options: {
```
elie222: `readEmail` may be better

apps/web/utils/gmail/label.ts
```diff
@@ -15,3 +15,33 @@ export async function labelThread(options: {
...
+  return gmail.users.threads.modify({
+    userId: "me",
+    id: threadId,
+    requestBody: {
+      removeLabelIds: ["UNREAD"],
+    },
+  });
+}
+
+export async function unMarkReadLabel(options: {
```
elie222: `markEmailUnread`

`unMark` is wrong as is

apps/web/utils/actions-client.ts
```diff
@@ -37,3 +39,22 @@ export async function onTrashThread(threadId: string) {
     });
   }
 }
+
+export async function onMarkUnmarkReadAction(
```
elie222: The naming is a bit all over the place here. `onMarkUnmarkReadAction`

It's either mark and unmark. But simpler might just be: `onMarkReadAction`

apps/web/utils/actions.ts
```diff
@@ -24,13 +24,15 @@ import { deletePosthogUser } from "@/utils/posthog";
...
 import { trashThread } from "@/utils/gmail/trash";
+import { markReadLabel, unMarkReadLabel } from "@/utils/gmail/label";
 import { env } from "@/env.mjs";
 import { isPremium } from "@/utils/premium";
 import {
   cancelUserPremium,
   upgradeUserToPremium,
 } from "@/utils/premium/server";
 import { ChangePremiumStatusOptions } from "@/app/(app)/admin/validation";
+import { threadId } from "worker_threads";
```
elie222: Bad import

apps/web/components/email-list/EmailPanel.tsx
```diff
@@ -29,6 +29,15 @@ export function EmailPanel(props: {
...
   const lastMessage = props.row.messages?.[props.row.messages.length - 1];
+  const [isRead, setIsRead] = useState(false);
+  const isUnread = useMemo(() => {
+    if (lastMessage?.labelIds?.includes("UNREAD")) {
+      setIsRead(false);
+    } else {
+      setIsRead(true);
+    }
+    return lastMessage?.labelIds?.includes("UNREAD");
+  }, [lastMessage?.labelIds]);
```
elie222: Don't think we need any extra state here. We were already marking rows as gray/white. See how that was done without state. We just get the row data passed in as props.

apps/web/components/ActionButtons.tsx
```diff
@@ -114,6 +133,31 @@ export function ActionButtons(props: {
...
+            onClick: onReadUnread,
+            icon: isReadUnreading ? (
+              <LoadingMiniSpinner />
+            ) : (
+              <MailCheckIcon
+                className="h-5 w-5 text-gray-700"
+                aria-hidden="true"
+              />
+            ),
+          },
```
elie222: Gmail uses the terminology "Mark as read" and "Mark as unread" - I'd use that so you don't have to reinvent the wheel here.

apps/web/components/ActionButtons.tsx
elie222: `isReadUnreading` is bad in any case.

`markingRead`, and `setMarkingRead` is better. Or:
`isMarkingRead`, and `setIsMarkingRead` is better.

elie222 (changes requested):
Thanks for the PR! Needs a bit of fixing.

Left a few comments about naming, misuse of React state and useMemo.

Also the most natural place to read in an email is when a row is clicked and a user has actually read the email.

Shalom2212:
> Thanks for the PR! Needs a bit of fixing.
> 
> Left a few comments about naming, misuse of React state and useMemo.
> 
> Also the most natural place to read in an email is when a row is clicked and a user has actually read the email.

Thank you For your comments, i'll fix it

Shalom2212:
I made all the changes as you mentioned please review it, if any changes needed please comment it!!
Thank you

PR #86, opened 2024-01-09
WIP: bulk action queue

elie222:
POC for better management of bulk tasks.

## Motivation

We can get rate limited on the OpenAI API or the Google API. Using `p-queue` we can limit bulk actions to 1 at a time. Ro 3 at a time. And make sure we're below any rate limit thresholds (and also retry failed tasks).

Would likely also clean up the codebase a bunch. We could use the same mechanism for bulk and single actions.

## Usage

Would be used like so:

```tsx
const queue = useQueue()

<button
  onClick={() => 
    queue.archiveEmails(['thread1', 'thread2']) // could also just be one thread if it's not a bulk action
  }>
  Archive
</button>
```

## Loading State

Probably also want to store the state of a task.
So that we know that thread1 and thread2 are `isArchiving: true`.
So if we need to show a spinner for a certain row in the UI we can do that.
A simple Zustand store might work for this. Reason we'd use Zustand over React Context is that only specific email rows would need to update. With React Context everything would update whenever the store changes.
Although could maybe deal with this optimisation later too, and just have a simple useState for it to begin with.

## Calling Gmail from Client vs Server

ATM, we're calling our server which calls the Gmail API. But really we could call the Gmail API directly if the client has the access token. It's inefficient to call the server API and we'd get faster responses if we called Gmail directly. For archiving an email there's no reason for it to route through our backend.
The change will likely be minimal as this code calls server functions, and calling client actions will be the same thing. But I put this here as part of the larger direction to keep in mind. Even if not implementing it today.


apps/web/app/(app)/providers.tsx
```diff
@@ -3,13 +3,16 @@ import React from "react";
...
+import { QueueProvider } from "@/providers/QueueProvider";
 
 export default function Providers(props: { children: React.ReactNode }) {
   return (
     <SWRProvider>
       {/* <GmailProvider> */}
       <SessionProvider>
-        <StatLoaderProvider>{props.children}</StatLoaderProvider>
+        <StatLoaderProvider>
+          <QueueProvider>{props.children}</QueueProvider>
```
Ipriyankrajai, line 14: Hey @elie222 ,

Do you think it is Ideal to Wrap QueueProvied in Root?

Because we need archiveEmails, readEmails, labelEmails, deleteEmails functionality into a single page that is `/mail`. So I think we should use QueueProvied inside `/mail`  page to make our app less bulk. P.S just a suggestion

elie222: Hey,

I missed this before. We actually use it in a few places. You can see emails in many different pages in the app so it makes sense to keep it at the `(app)/layout` level.

elie222:
Closing this PR as @Ipriyankrajai opened another one for it.

PR #87, opened 2024-01-10
POC for Bulk archive using Action queue

Ipriyankrajai:
Previous conversation here: 
https://github.com/elie222/inbox-zero/pull/86

POC for better management of bulk tasks.

## Motivation

We can get rate limited on the OpenAI API or the Google API. Using `p-queue` we can limit bulk actions to 1 at a time. Ro 3 at a time. And make sure we're below any rate limit thresholds (and also retry failed tasks).

Would likely also clean up the codebase a bunch. We could use the same mechanism for bulk and single actions.

## Usage

Would be used like so:

```tsx
const queue = useQueue()

<button
  onClick={() => 
    queue.archiveEmails(['thread1', 'thread2']) // could also just be one thread if it's not a bulk action
  }>
  Archive
</button>
```

## Loading State

Probably also want to store the state of a task.
So that we know that thread1 and thread2 are `isArchiving: true`.
So if we need to show a spinner for a certain row in the UI we can do that.
A simple Zustand store might work for this. Reason we'd use Zustand over React Context is that only specific email rows would need to update. With React Context everything would update whenever the store changes.
Although could maybe deal with this optimisation later too, and just have a simple useState for it to begin with.

## Calling Gmail from Client vs Server

ATM, we're calling our server which calls the Gmail API. But really we could call the Gmail API directly if the client has the access token. It's inefficient to call the server API and we'd get faster responses if we called Gmail directly. For archiving an email there's no reason for it to route through our backend.
The change will likely be minimal as this code calls server functions, and calling client actions will be the same thing. But I put this here as part of the larger direction to keep in mind. Even if not implementing it today.


apps/web/components/email-list/EmailList.tsx
```diff
@@ -369,11 +370,15 @@ export function EmailList(props: {
   const onArchiveBulk = useCallback(async () => {
     toast.promise(
       async () => {
+        const listOfSelectedThreadId = [];
```
elie222: I prefer to use filter here. Instead of for + push.

elie222 (changes requested):

apps/web/components/email-list/EmailList.tsx
Ipriyankrajai: How do you suggest we manage a `setIsArchiving((s) => ({ ...s, [threadId]: true }));` Archiving loader?

elie222:
Merge conflict which I fixed here: https://github.com/elie222/inbox-zero/pull/86

elie222:
Ah. I see it's only sitemap now

elie222:
Merged here:
https://github.com/elie222/inbox-zero/pull/103

Issue #54, opened 2023-12-27
Store data in browser with Indexeddb


elie222:
We can store gigabytes of data in an indexddb in the browser.
Use `idb` package to do this.

The advantage here is extra privacy for users. After this is done we can remove large parts of the backend. The client can load the data itself and use the Gmail API to archive/label/reply.
It will also mean faster loading times.

The core tables to use are:
* Emails
* Labels

We may need more tables after for sorting in different ways, or to help with analytics.

sharma-shanu:
@elie222 , Can you assign this task to me ?

elie222:
Not relevant atm.

PR #97, opened 2024-01-22
Caching and handling Indexeddb operations in Service Workers 

sharma-shanu:
Issue Link :- #54 

This PR implements following features :-

1. Adds a new service worker for the web app (apps/web).
2. Uses worker thread to perform all indexeddb related ops.
3. Implements caching of images on the browser for faster load.
4. Implements analytics for user emails.


sharma-shanu:
This also helps in removing complexity of indexeddb related transaction from core web application, so the web app can remain oblivious to the complexity of making and handling request network calls.

elie222:
> This also helps in removing complexity of indexeddb related transaction from core web application, so the web app can remain oblivious to the complexity of making and handling request network calls.

Great stuff. I see I've caused some merge conflicts.

What are the steps needed to push this to completion?

sharma-shanu:
> > This also helps in removing complexity of indexeddb related transaction from core web application, so the web app can remain oblivious to the complexity of making and handling request network calls.
> 
> Great stuff. I see I've caused some merge conflicts.
> 
> What are the steps needed to push this to completion?

Hey Elie 👋🏻, 

This is mostly done, I am ironing out some of the functionalities.

There are 2 features i want to add to this :- 

1. During signup/initial login, when the user asks for stats, we load the data into tinybird in the background of next server, during this we currently show no data is available, while the data is being loaded in the background (this takes about a minute or two) . Currently, we have an iterative solution which is taking longer imo. I want to research on a solution in gmail api where we can make all the necessary requests simultaneously. 

2. When a user logs out, I want to clear the indexeddb objectStores so that snooping is not available on stale emails.

elie222:
> 1. During signup/initial login, when the user asks for stats, we load the data into tinybird in the background of next server, during this we currently show no data is available, while the data is being loaded in the background (this takes about a minute or two) . Currently, we have an iterative solution which is taking longer imo. I want to research on a solution in gmail api where we can make all the necessary requests simultaneously.

I agree this isn't great and needs fixing. But I'd keep it for another PR. I'd prefer to keep this one focused.

> 2. When a user logs out, I want to clear the indexeddb objectStores so that snooping is not available on stale emails.

Sure. I'm guessing this is a one liner in the log out function?

apps/web/components/WithServiceWorker.tsx
```diff
@@ -0,0 +1,43 @@
...
+    [],
+  );
+
+  useEffect(() => {
+    if (message && cb) {
+      if (actions && actions.length && actions.includes(message.type))
+        cb(message);
+      else cb(message);
+    }
+  }, [message, cb, actions]);
```
elie222: 1. Why is this split into two useEffects?

Wouldn't this be simpler? And then you don't need `useState`?

```ts
useEffect(
    () =>
      navigator.serviceWorker.addEventListener(
        "message",
        (event: MessageEvent<Message>) => {
          // logic here...
          cb(message);
        }
      ),
    [],
  );
```

2. Also wouldn't it be easier to just turn the above into a hook `useServiceWorker` at that point?

3. Would it make sense to put this at the app/(app)/layout level and just call it once?

apps/web/app/layout.tsx
```diff
@@ -67,6 +68,7 @@ export default function RootLayout({
             <PostHogPageview />
           </Suspense>
           <Providers>{children}</Providers>
+          <Script src="/service-worker.js" strategy="beforeInteractive" />
```
elie222, line 73: Does it make sense to put this in the app specific layout instead?

`app/(app)/layout`

apps/web/app/api/user/stats/tinybird/load/validation.ts
```diff
@@ -2,5 +2,7 @@ import { z } from "zod";
 
 export const loadTinybirdEmailsBody = z.object({
   loadBefore: z.coerce.boolean().optional(),
+  timestamp: z.coerce.number().optional(),
 });
 export type LoadTinybirdEmailsBody = z.infer<typeof loadTinybirdEmailsBody>;
+export type LoadIDBEmailsBody = z.infer<typeof loadTinybirdEmailsBody>;
```
elie222: Not sure it makes sense to rename this.
So something isn't consistent here.

apps/web/public/sw.js
```diff
@@ -0,0 +1,384 @@
...
+
+  //cleaning (if we want zeros, we can remove cleaning)
+  clusters.forEach((ele) => {
+    const keys = Object.keys(ele);
+    for (key of keys) if (ele[key] === 0) delete ele[key];
+  });
+
+  return {
+    result: clusters,
+    allCount: clusters.reduce((sum, ele) => sum + (ele.All ?? 0), 0),
```
elie222, line 374: `sumBy` is cleaner from lodash. We use the package anyway in other places

packages/tinybird/src/publish.ts
```diff
@@ -30,6 +30,7 @@ const tinybirdEmail = z.object({
   sizeEstimate: z.number().nullish(), // Estimated size in bytes
 });
 export type TinybirdEmail = z.infer<typeof tinybirdEmail>;
+export type IndexedDBEmail = z.infer<typeof tinybirdEmail>;
```
elie222: This is the tinybird package and has nothing to do with indexeddb. Please undo this change.

elie222 (changes requested):
Hey,

Took a look through this. I'd like to avoid any dependency between Tinybird and IDB.
They're unrelated. The data in IDB can be loaded directly from Gmail. It doesn't need to touch Tinybird at all. In the future I may look to remove Tinybird completely, and having it mixed into IDB will make this harder. Having IDB as part of the data store was a way to move away from Tinybird. Not to more tightly couple them together.

I'd love to keep changes much smaller and not impact the entire app.
So picking one specific spot to add it. We can expand after that.

One really small lift would be to add Labels to the mail client, somewhat similar to this:
https://ui.shadcn.com/examples/mail (but use our current UI) for the sidenav.

1. Load labels from gmail
2. Store in IDB
3. Display labels when Mail tab is open, fetching the data from IDB, or a spinner if still loading.

And then clear data on log out.

Another step where this could be done after that (in a new PR so we don't mix things up):

Load email data when the app starts up. Direct from gmail. Don't involve Tinybird at all. And then poll every 10s if we need new data. This should be done at the (app)/layout level so it isn't loaded multiple times.

And then choose just one place to use it.
eg on the /stats page for `Who emails you most`.
We can expand its usage to other places in the app after that. But I'd really want to keep the changes somewhat separate.

sharma-shanu:
> > 1. During signup/initial login, when the user asks for stats, we load the data into tinybird in the background of next server, during this we currently show no data is available, while the data is being loaded in the background (this takes about a minute or two) . Currently, we have an iterative solution which is taking longer imo. I want to research on a solution in gmail api where we can make all the necessary requests simultaneously.
> 
> I agree this isn't great and needs fixing. But I'd keep it for another PR. I'd prefer to keep this one focused.
> 
> > 2. When a user logs out, I want to clear the indexeddb objectStores so that snooping is not available on stale emails.
> 
> Sure. I'm guessing this is a one liner in the log out function?

Yes, it would be a one liner.

And i will create a separate PR for converting iterative solution.

apps/web/public/sw.js
sharma-shanu: I agree, I wanted to use typescript and npm packages inside the service worker, but unfortunately i could not use imports and exports within it,

`Uncaught SyntaxError: Cannot use import statement outside a module `;
`sw.js:4 Uncaught ReferenceError: exports is not defined` even when i was not exporting anything from service worker.

packages/tinybird/src/publish.ts
sharma-shanu: Addressed.

apps/web/app/api/user/stats/tinybird/load/validation.ts
sharma-shanu: Addressed.

apps/web/app/layout.tsx
sharma-shanu: I think it makes sense. Service worker is registered fully (installed, activated and subscribed) by the time user is authenticated when imported from the root layout. 

Can I create a separate PR for it as it would need more testing for registration ?

apps/web/components/WithServiceWorker.tsx
sharma-shanu: 1,2. Addressed.

3. We can now use hook 'useServiceWorker()' to subscribe to one action, list of action types and also all the messages coming from service worker.

sharma-shanu:
https://github.com/elie222/inbox-zero/pull/97#pullrequestreview-1852227372

Hey @elie222 👋🏻, hope you are doing great.

I just made few changes to the sidenav and added labels section to it.

Here are screenshots for it:- 
Screenshot 2024-02-02 at 6 20 14 PM

https://github.com/elie222/inbox-zero/assets/85648738/021e1901-886b-46ce-89a9-ddf5c2d3408a

When user has more than 4 labels, it will convert to expandable visibility toggle. Otherwise all options will be visible.

Let me know your opinion on how do they look and do they meet what you had in mind.

Thank you so much 😁.

apps/web/app/(app)/idb-playground/page.tsx
```diff
@@ -0,0 +1,76 @@
+"use client";
```
sharma-shanu, line 1: This route is mostly dev route created to test changes for indexeddb migration.

apps/web/app/api/user/stats/emails/all/get-all.ts
```diff
@@ -0,0 +1,207 @@
+import {
```
sharma-shanu, line 1: This api fetches all the email data from google client (optionally based on timestamp ), similar to https://github.com/elie222/inbox-zero/blob/main/apps/web/app/api/user/stats/tinybird/load/load-emails.ts.

apps/web/app/api/user/stats/emails/sw/route.ts
```diff
@@ -0,0 +1,16 @@
+import { NextResponse } from "next/server";
```
sharma-shanu, line 1: This api route helps in returning an empty message response iff service working is not installed on the browser or fails to intercept a fetch request.

apps/web/components/SideNav.tsx
```diff
@@ -190,6 +196,61 @@ export function SideNav(props: {
...
+        {label.name}
+      </a>
+    ));
+  }
+
+  const LabelSection = (
+    <>{tagItem(data?.labels?.filter((label) => label?.type === "user"))}</>
+  );
+
+  const LabelSectionExpandable = (
```
sharma-shanu: This is rendered on the UI, if the user has more than 4 "user" created labels in their account as an expandable section, similar to GMail's UI.

apps/web/public/service-worker.js
```diff
@@ -0,0 +1,12 @@
+function registerServiceWorker() {
```
sharma-shanu, line 1: This registers the service worker on the browser.

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
+/* Service Worker */
```
sharma-shanu, line 1: This is the implementation of service worker which intercepts all fetch/network requests coming from browser and acts as a proxy between client and server/cached responses.

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+  hour: 3600_000,
+  day: 86400_000,
+  week: 604800_000,
+  month: 2629743_000,
+  year: 31556926_000,
+};
+
+self.addEventListener("install", (event) => {
+  console.log(`Version ${version} installed`);
+  self.skipWaiting();
```
sharma-shanu, line 36: Immediately activates service worker upon installation.

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+});
+
+self.addEventListener("activate", (event) => {
+  //service worker is activated
+  event.waitUntil(Promise.resolve().then(openDB));
```
sharma-shanu, line 40: This opens up a new connection for the indexedDB as soon as the service worker is activated on the broswer.

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+  console.log("activated");
+});
+
+self.addEventListener("fetch", (event) => {
+  //service worker intercepted a fetch call
+  console.log(event.request.url);
+  if (!isAuth(event.request.url))
```
sharma-shanu, line 47: skips all the auth request and callback request coming from client and gmail server.

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+      resolve();
+    };
+  });
+};
+
+async function handleFetch(request) {
+  const { url } = request;
+  if (url.includes("/api/google/labels")) {
+    // this should be called regardless.
+    fetchAndUpdateIDB(request.clone());
```
sharma-shanu: This function fetches new/updated labels from server in the background and updates indexeddb and client with them.

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+        console.log("Error while getting data from indexeddb", error.message);
+        // return new Response(`Error:  + ${error.message}`, {
+        //   status: 500,
+        //   headers: { "Content-Type": "text/plain" },
+        // });
+      }
+    }
+  }
+
+  if (
```
sharma-shanu, line 117: This caches all the static images in the browser.

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+
+  // open the database connection
+  const data = await clonedResponse.json();
+  if (!data?.labels) return;
+
+  await saveLabels(data.labels);
+  messageClients(MESSAGES.LABELS_UPDATED, data);
+}
+
+function messageClients(type, payload) {
```
sharma-shanu, line 196: Broadcasts data to all the browser tabs linked to the service worker with the updated data.

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+
+// all inclusive
+const numberOfClusters = (start, end, size) => {
+  const window = (end - start + 1) / size;
+  return Math.floor(window) + (window - Math.floor(window) > 0 ? 1 : 0);
+};
+
+const getClusteredArray = (start, end, size) => {
+  const nClusters = numberOfClusters(start, end, size);
+  const clusters = [];
```
sharma-shanu, line 318: A cluster is a period of time with all the stats about all activity during that period. It could be of size day/week/month/year.

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+  return {
+    result: clusters,
+    allCount: clusters.reduce((sum, ele) => sum + (ele.All ?? 0), 0),
+    inboxCount: clusters.reduce((sum, ele) => sum + (ele.Unarchived ?? 0), 0),
+    readCount: clusters.reduce((sum, ele) => sum + (ele.Read ?? 0), 0),
+    sentCount: clusters.reduce((sum, ele) => sum + (ele.Sent ?? 0), 0),
+  };
+}
+
+async function clearAllData() {
```
sharma-shanu, line 381: This function clears all the data upon logout by the user.

apps/web/components/SideNav.tsx
sharma-shanu: https://github.com/elie222/inbox-zero/pull/97#issuecomment-1923793958

elie222:
How's it going here?
I hope to add a better view for the mail client today/tomorrow. And that could be a helpful point to add a small PR for the labels only.

sharma-shanu:
> How's it going here? I hope to add a better view for the mail client today/tomorrow. And that could be a helpful point to add a small PR for the labels only.

Sure @elie222, 

I have added the labels view to sidenav in this PR. 

I saw the new changes on discord and they look nice.

I can create a small PR just for labels once the new view is up and merged.

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
+/* Service Worker */
+
+const version = 1;
+const dbVersion = 2;
+const database = "inbox-zero";
+let DB = null;
+let dynamicName = `dynamicCache`;
+let IMAGE_CACHE = "IMAGE_CACHE";
```
elie222: Unused

elie222: prefer `const` to `let`. Maybe I enable this as an ESLint warning

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+const LABEL_STORE = {
+  name: "labels",
+  key: "id",
+  indexes: [],
+};
+const EMAIL_STORE = {
+  name: "emails",
+  key: "gmailMessageId",
+  indexes: [{ name: "timestampIDX", property: "timestamp", params: {} }],
```
elie222: not sure about naming conventions here, but what about just calling it `timestamp`? Is IDX a convention?

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+const TIME_CONSTANT = {
+  // in ms
+  hour: 3600_000,
+  day: 86400_000,
+  week: 604800_000,
+  month: 2629743_000,
+  year: 31556926_000,
+};
+
+self.addEventListener("install", (event) => {
```
elie222: can make this `_event`

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+});
+
+self.addEventListener("fetch", (event) => {
+  //service worker intercepted a fetch call
+  console.log(event.request.url);
+  if (!isAuth(event.request.url))
+    event.respondWith(
+      handleFetch(event.request).then((response) => {
+        // TODO: Remove this
+        if (event.request.url.includes("tinybird")) {
```
elie222: what's this TODO?

apps/web/public/sw.js
elie222: wouldn't this be simpler?
event.waitUntil(openDB());

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+  year: 31556926_000,
+};
+
+self.addEventListener("install", (event) => {
+  console.log(`Version ${version} installed`);
+  self.skipWaiting();
+});
+
+self.addEventListener("activate", (event) => {
+  //service worker is activated
```
elie222: comments are okay, but this seems self explanatory

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+        // TODO: Remove this
+        if (event.request.url.includes("tinybird")) {
+          console.log(response);
+        }
+        return response;
+      }),
+    );
+});
+
+self.addEventListener("message", (event) => {
```
elie222: do we need this? or do you want to comment it out?

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+  event.waitUntil(Promise.resolve().then(openDB));
+  console.log("activated");
+});
+
+self.addEventListener("fetch", (event) => {
+  //service worker intercepted a fetch call
+  console.log(event.request.url);
+  if (!isAuth(event.request.url))
+    event.respondWith(
+      handleFetch(event.request).then((response) => {
```
elie222: I prefer async/await to match the rest of the codebase

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+  self.skipWaiting();
+});
+
+self.addEventListener("activate", (event) => {
+  //service worker is activated
+  event.waitUntil(Promise.resolve().then(openDB));
+  console.log("activated");
+});
+
+self.addEventListener("fetch", (event) => {
```
elie222, line 44: why not just have an if that focuses on the specific requests we want to cache?

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+        return response;
+      }),
+    );
+});
+
+self.addEventListener("message", (event) => {
+  //message from webpage
+});
+
+const openDB = (callback) => {
```
elie222: do we ever pass a callback?

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+      }
+      resolve();
+    };
+  });
+};
+
+async function handleFetch(request) {
+  const { url } = request;
+  if (url.includes("/api/google/labels")) {
+    // this should be called regardless.
```
elie222: why? could also just remove the comment?

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+    return new Response(
+      JSON.stringify({
+        message: "No stored data found in indexedDB for analytics",
+      }),
+    );
+  }
+  return fetch(request);
+}
+
+async function fetchAndUpdateIDB(request) {
```
elie222: this has a generic name but is also handling things related to labels?

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
+/* Service Worker */
+
+const version = 1;
```
elie222, line 3: can we use the `idb` package in the webworker file?

apps/web/utils/indexeddb/index.ts
```diff
@@ -0,0 +1,31 @@
+"use client";
+
+import { openDB, IDBPDatabase } from "idb";
```
elie222: is this still in use?

apps/web/public/sw.js
elie222: IMAGE_CACHE doesn't seem to be in use

apps/web/public/sw.js
```diff
@@ -0,0 +1,411 @@
...
+    url.includes(".png") ||
+    url.includes(".jpeg")
+  ) {
+    const cacheResponse = await caches.match(request);
+    if (cacheResponse) return cacheResponse;
+    const fetchResponse = await fetch(request);
+
+    //save in image cache
+    console.log(`save an IMAGE file ${url}`);
+    const cache = await caches.open(IMAGE_CACHE);
```
elie222, line 129: we save in the image cache, but when do we ever read from it?

elie222:
Hey,

Went through a bunch of it right now. I didn't manage to cover everything. Would be really nice to have this broken out into separate PRs. Would be easier to merge that way. Because it covers a lot of functionality right now.

When testing the app locally, I didn't notice any speed improvements. Should I have seen instant loading of labels in the sidebar for example?

apps/web/public/sw.js
sharma-shanu: https://github.com/elie222/inbox-zero/pull/97/files#diff-f50be0345a656c79db8f461fd7398526a2706e339548490178d905c062d3feedR140-R141

sharma-shanu: https://github.com/sharma-shanu/inbox-zero/blob/feat/indexdb-with-service-workers/apps/web/public/sw.js#L126-L127

apps/web/utils/indexeddb/index.ts
sharma-shanu: No @elie222 , these are not in use. I think these came to this PR when i created a new branch from idb-playground branch.

apps/web/public/sw.js
sharma-shanu: Can you explain it a bit more?

apps/web/public/sw.js
sharma-shanu: `const cacheResponse = await caches.match(request);`
   ` if (cacheResponse) return cacheResponse;`

apps/web/app/api/user/stats/emails/all/get-all.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  // 2. fetch each email and return them --
+  const messages = await getMessagesBatch(
+    res.data.messages?.map((m) => m.id!) || [],
+    accessToken,
+  );
+
+  const emailsToPublish: IndexedDBEmail[] = (
+    await Promise.all(
+      // why is this function async ?? This could simply be sync (including promise)
```
elie222, line 158: good point. i fixed this in `main` now

elie222: i'm guessing it happened because async was needed here in the past. but isn't in the current version of the code.

apps/web/app/api/user/stats/emails/all/get-all.ts
elie222: seems like there's a lot of copy paste from the other file. There wasn't an option to create a new function and share code?

Issue #89, opened 2024-01-18
Allow using regular Redis


elie222:
A solution is to update the Docker Compose to this:
https://github.com/hiett/serverless-redis-http#via-docker-compose

Then you don't need to use Upstash Redis.

elie222:
*Not doing this. The above method should work*

Another option is to support it in the code like so:

```ts
import { env } from "@/env.mjs";

// Function to dynamically import Redis based on user preference or environment variable
async function getRedisClient() {
  let RedisClient;
  if (env.UPSTASH_REDIS_URL) {
    const { Redis } = await import("@upstash/redis");
    RedisClient = new Redis({
      url: env.UPSTASH_REDIS_URL,
      token: env.UPSTASH_REDIS_TOKEN,
    });
  } else {
    const { createClient } = await import("redis");
    RedisClient = createClient({ url: env.REDIS_URL });
    await RedisClient.connect();
  }
  return RedisClient;
}

export const redis = await getRedisClient();
```

And change `target` to `ES2017` in `packages/tsconfig/nextjs.json`.

Also needs some TS work to make it compatible. Here's a POC:
https://github.com/elie222/inbox-zero/tree/redis-without-upstash

But I think I prefer the Docker option for now so we don't need to install two different packages.

asif-reh:
Hi @elie222 I would love to solve this issue could you please assign me.

elie222:
> Hi @elie222 I would love to solve this issue could you please assign me.

Done!

It may be very simple and just copy pasting a few lines into the Docker Compose. Just need to check that it actually works.
I don't want to mess around with changing our code (the 2nd option I put above), as it will be more long term maintenance.

PR #101, opened 2024-01-30
feat: adding serverless-redis-http to docker compose

Ipriyankrajai:
fixes: #89

P.S. @elie222 rename the env variable UPSTASH_REDIS_URL, UPSTASH_REDIS_TOKEN to REDIS_URL, REDIS_TOKEN respectively.


docker-compose.yml
```diff
@@ -20,6 +20,14 @@ services:
       - 6380:6379
     volumes:
       - database-data:/data
+  serverless-redis-http:
+    ports:
+      - "8079:80"
+    image: hiett/serverless-redis-http:latest
+    environment:
+      SRH_MODE: env
+      SRH_TOKEN: "27191372ba5c50fd95f59baa43d0e3d2"
```
elie222: What’s this value? Should it be hard coded?

apps/web/.env.example
```diff
@@ -10,8 +10,9 @@ GOOGLE_CLIENT_SECRET=
 
 OPENAI_API_KEY=
 
-UPSTASH_REDIS_URL=redis://localhost:6380
-UPSTASH_REDIS_TOKEN=''
+# local redis config
+REDIS_URL="http://localhost:8079"
+REDIS_TOKEN="27191372ba5c50fd95f59baa43d0e3d2"
```
elie222: Did you mean to commit this?

elie222 (changes requested):

docker-compose.yml
Dharmik6898: No, we need a random secret key so I generated it from  https://generate-secret.vercel.app/32

apps/web/.env.example
Dharmik6898: Yes, we need to set redis_url and redis_token in the .env.example so that when the user 
`cp apps/web/.env.example apps/web/.env` they will automatically have the Redis URL and token set up for their local environment without configuring an Upstash account.

docker-compose.yml
Ipriyankrajai: We need a random secret key so I generated it from https://generate-secret.vercel.app/32

apps/web/.env.example
Ipriyankrajai: Yes, we need to set redis_url and redis_token in the .env.example so that when the user cp apps/web/.env.example apps/web/.env they will automatically have the Redis URL and token set up for their local environment without configuring an Upstash account.

docker-compose.yml
elie222: You can do this to use the env already set for it. env_file tells it where to find it. also add a comment in the .env file what token is supposed to be now that you've removed the word upstash from it:

serverless-redis-http:
    ports:
      - "8079:80"
    image: hiett/serverless-redis-http:latest
    env_file:
      - ./apps/web/.env
    environment:
      SRH_MODE: env
      SRH_TOKEN: ${SRH_TOKEN}
      SRH_CONNECTION_STRING: "redis://redis:6379"

elie222: You can add a comment about generate-secret.vercel.app/32 in the .env too

elie222: CleanShot 2024-01-30 at 23 50 51@2x
I'd probably keep it consistent with this. Because someone might accidentally use your generated secret key instead of creating their own. And that's more dangerous

Ipriyankrajai: Yes, we can use env. but issue is our docker-compose is in root directory and env is in /apps/web. I went through few documentation and it says we cant use env from other directory [stack overflow issue](https://stackoverflow.com/questions/56222897/how-to-use-environment-variables-with-docker-compose-from-other-directory)

Ipriyankrajai: Sure, will add that. I went through documentation and was not able to find any solution [here](https://stackoverflow.com/questions/56222897/how-to-use-environment-variables-with-docker-compose-from-other-directory)

Ipriyankrajai:
Hey @elie222,

I have made the changes

elie222:
Looks good! Do you mind just switching back the env var names to `UPSTASH_`.
I like what you did. But this change will break the app for anyone using it. And I'll have to update env vars myself too.

I may make this change after incrementally but now will be best to just leave the name as is if that's okay.

Also fix the pnpm.lock conflict we have 😢 .

Ipriyankrajai:
Done @elie222.

Issue #116, opened 2024-02-07
Welcome to email analytics POP UP opening everytime 


ketanMehtaa:
![image](https://github.com/elie222/inbox-zero/assets/45426198/b1320f2f-c208-4af9-be37-065b1b7c1063)

PR #118, opened 2024-02-08
fix(stats/StatsOnboarding): with a version bump to usehooks-ts@2.13.0

sharma-shanu:
Issue :- #116 

This PR fixes the issue by bumping up the version of usehooks-ts library (from 2.10.0 to 2.13.0 and above).

The utility hook "useLocalStorage" had a bug where it returned the "initialValue" passed irrespective of whether the field is available or not in the localStorage. 

Hence, making it difficult to use a sentinel value in visibility logic [(StatsOnboarding.tsx).](https://github.com/elie222/inbox-zero/blob/main/apps/web/app/(app)/stats/StatsOnboarding.tsx#L19-#L29)

This has been patched in version 2.13.0.


elie222:
Nice! Noticed that the modal kept popping up recently. Thanks for the fix.

Issue #125, opened 2024-02-10
Issue with mail navigation


Ipriyankrajai:
https://github.com/elie222/inbox-zero/assets/32142351/60e84673-bd36-4026-abbd-75ca03326e32

PR #126, opened 2024-02-10
Fix: mail navigation

Ipriyankrajai:
Fix: #125

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,6 @@ The external services that are required are:
 - [OpenAI](https://platform.openai.com/api-keys)
 - [Google OAuth](https://console.cloud.google.com/apis/credentials)
 - [Google PubSub](https://console.cloud.google.com/cloudpubsub/topic/list) - see set up instructions below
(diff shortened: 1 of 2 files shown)


Ipriyankrajai:
Fix: #127

elie222:
This changes README too (and not sure in ways it needs to be changed).

For the bug itself, where does it happen to you? Because it links me to these pages correctly: https://www.getinboxzero.com/mail?type=sent

Ipriyankrajai:
First go to compose page than then try to go to mail or draft page

Ipriyankrajai:
By changing readme i tried to Fix: #127

elie222:
> First go to compose page than then try to go to mail or draft page

Ah okay. I actually got rid of compose page now. For multiple changes it would be better to have it as two different PRs

elie222:
Also about the README, Redis is something you want to set if you're deploying it. Db url is too. So maybe I should make that more clear.

PR #142, opened 2024-02-19
Add a service worker.

sharma-shanu:
Issue :- #54 

This PR adds :- 

1. A basic service worker at the top level app layout.
2. Creates indexeddb schema with specified indexes and a connection object to the database.

Screenshot 2024-02-20 at 2 19 42 AM

Service worker is registered at the top level of app to ensure installation, activation and registration as soon as user uses the app.

This ensures :- 
- Service worker is running before the user logs into the app.
- The db connection object is available readily to respond with cached information since the first page of app.
- Since there is only one service worker allowed per scope, a root level service worker can be used to cache static resources from the very first page.

### DIFF
--- a/apps/web/app/layout.tsx
+++ b/apps/web/app/layout.tsx
@@ -9,6 +9,7 @@ import { Inter } from "next/font/google";
 import localFont from "next/font/local";
 import { env } from "@/env.mjs";
 import Providers from "@/app/(app)/providers";
+import Script from "next/script";
 
 const inter = Inter({
   subsets: ["latin"],
@@ -72,6 +73,7 @@ export default function RootLayout({
         </PostHogProvider>
         <SpeedInsights />
         <Analytics />
+        <Script src="/service-worker.js" strategy="beforeInteractive" />
         {env.NEXT_PUBLIC_GTM_ID ? (
           <GoogleTagManager gtmId={env.NEXT_PUBLIC_GTM_ID} />
         ) : null}
--- a/apps/web/public/service-worker.js
+++ b/apps/web/public/service-worker.js
@@ -1,0 +1,13 @@
+function registerServiceWorker() {
+    console.log("registerServiceWorker fired");
+    if (typeof window !== "undefined") {
+      if ("serviceWorker" in navigator) {
+        navigator.serviceWorker.register("/sw.js").then((registration) => {
+          console.log("Service Worker registration successful:", registration);
+        });
+      }
+    }
+  }
+  
+  registerServiceWorker();
+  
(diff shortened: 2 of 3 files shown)


elie222:
> A basic service worker at the top level app layout.

Why? What benefit does this add to the app?

> Creates indexeddb schema with specified indexes and a connection object to the database.

But it isn't used at all. So it also doesn't add anything.

### More notes

You can add `idb` without Service Workers (and vice versa). But more important is **why** we're adding them. Adding more tech for the sake of it doesn't help us. And adds complexity to the codebase. And caching/invalidation of cache has its challenges too.

Example use cases would be:

* Image loading is slow. Let's cache them.
* We want the app to work offline. Let's cache responses with a service worker.

But neither of these are challenges we're facing in the app right now. So I don't want to tackle either of them.

### What is the actual task we want to accomplish and how do we get there

* Load labels - easiest way to do this is to just request labels from Gmail. Premature optimisation right now to have it cached. The challenge with caching it is that we then have to also check when the labels are updated and update our cache. Easiest is to just call the labels when needed.

* Replace Tinybird with Indexeddb - the benefit here is for those that don't want data stored in Tinybird, and also makes it easier for those deciding to self-host. For the typical end user this won't make a huge difference. A challenge here will be whether we can get the same performance because Tinybird is very fast. Queries take 10ms even on large datasets.

If this were to be implemented this is how it would happen:

* App loads up
* Load data from Gmail into the indexeddb. The API request for this would be very simple. It's just a gmail.threads.list setting the date range each time to be further back. It doesn't have to be like the code in this PR: https://github.com/elie222/inbox-zero/pull/97
* Extract relevant data like unsubscribe links from emails. This will also happen in the user's browser just before storing the data.
* We'd then calculate the data needed on the client. eg. all newsletter senders. Hopefully there wouldn't be any slowness here compared to tb, even if we've stored a lot of emails.
* Update cache: as new emails come in we'd have to update indexeddb store. This can happen when the user logs in.

### Final thoughts

I'm wary of making these changes atm. Because of the challenges around caching and performance. And how much benefit there really is.

sharma-shanu:
> You can add idb without Service Workers (and vice versa). But more important is why we're adding them. Adding more tech for the sake of it doesn't help us. And adds complexity to the codebase. And caching/invalidation of cache has its challenges too.

I understand. 

The main idea behind creating a service worker was to :-

- move logical complexity away from the main frontend app.
- ensure the app works in the same way as it already did without any change.
- making the main app a bit more resilient since some modern browser clear idb in 7-14 days.
- minor performance improvements.

For now I've closed the existing PRs.

Learnt a lot while working on these, thank you so much reviewing them.

elie222:
Also a learning curve for me how to manage these PRs. And hopefully I can manage them better in the future. I need to be more clear on the requirements and the "why" for any code we add.

Issue #123, opened 2024-02-09
Don't show congrats on reaching inbox zero when not on inbox tab


elie222:
This shouldn't show for Drafts for example:
![CleanShot 2024-02-09 at 12 56 28@2x](https://github.com/elie222/inbox-zero/assets/3090527/b517b896-d8f6-4d2f-b8a2-e35a2aacffa9)

ketanMehtaa:
What should I show when the draft (or anything other than the inbox) is empty?

elie222:
> What should I show when the draft (or anything other than the inbox) is empty?

"You don't have any saved drafts."

PR #147, opened 2024-02-28
message for empty draft , sent , archive handled.

ketanMehtaa:
#123

### DIFF
--- a/apps/web/components/Celebration.tsx
+++ b/apps/web/components/Celebration.tsx
@@ -4,17 +4,44 @@ import Image from "next/image";
 import { getCelebrationImage } from "@/utils/celebration";
 import { Button } from "@/components/Button";
 
-export function Celebration(props: { message: string }) {
+export function Celebration(props: { type: string }) {
   const [active, setActive] = useState(false);
 
   useEffect(() => {
     setActive(true);
   }, []);
 
+  if (props.type == "draft") {
+    return (
+      <>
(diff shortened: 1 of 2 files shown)


apps/web/components/Celebration.tsx
```diff
@@ -4,17 +4,44 @@ import Image from "next/image";
 import { getCelebrationImage } from "@/utils/celebration";
 import { Button } from "@/components/Button";
 
-export function Celebration(props: { message: string }) {
+export function Celebration(props: { type: string }) {
```
elie222, line 7: Thanks. I'd just use the existing format and pass in the relevant message. No need to change the props. Or this component at all.

apps/web/components/Celebration.tsx
```diff
@@ -4,17 +4,44 @@ import Image from "next/image";
...
   const [active, setActive] = useState(false);
 
   useEffect(() => {
     setActive(true);
   }, []);
 
+  if (props.type == "draft") {
+    return (
+      <>
+        <div className="mt-20 flex items-center justify-center font-cal text-2xl text-gray-900">
```
elie222: This style is copy passed each time. But either way, just pass in the message and don't edit this component.

elie222:
Thanks. There's no need to show a celebration at all for drafts. You can have a simple message without any graphic for drafts.

PR #159, opened 2024-03-06
feat: cancel subscription

Ipriyankrajai:
fixes: #153


apps/web/utils/actions.ts
```diff
@@ -106,6 +125,9 @@ export async function deleteAccountAction() {
     captureException(error);
   }
 
+  if (lemonSqueezySubscriptionItemId)
+    await cancelSubScriptionForUser(lemonSqueezySubscriptionItemId);
+
   await prisma.user.delete({ where: { email: session.user.email } });
```
Ipriyankrajai: I think we should delete the user after the subscription tenure is ended.

apps/web/utils/actions.ts
```diff
@@ -89,6 +92,22 @@ export async function deleteAccountAction() {
...
   if (!session?.user.email) throw new Error("Not logged in");
 
+  const user = await prisma.user.findUniqueOrThrow({
+    where: { id: session.user.id },
+    select: {
+      premium: {
+        select: {
+          id: true,
+          tier: true,
+          lemonSqueezySubscriptionItemId: true,
```
elie222: I don't think we need any of these.

We need: `lemonSqueezySubscriptionId` (which is different to `lemonSqueezySubscriptionItemId`)

apps/web/utils/actions.ts
elie222: No, don't delete the user. "leaving a user's data in Stripe after they delete their account is generally allowed under GDPR, provided that you have a legitimate reason for retaining this data, such as for financial records, compliance with tax law, or other regulatory requirements."

elie222 (changes requested):
Merge conflict too :(

apps/web/utils/actions.ts
Ipriyankrajai: Done

apps/web/utils/actions.ts
Ipriyankrajai: For now i have commented that part

apps/web/utils/actions.ts
```diff
@@ -92,6 +93,21 @@ export async function deleteAccountAction() {
   const session = await auth();
   if (!session?.user.email) throw new Error("Not logged in");
 
+  const user = await prisma.user.findUniqueOrThrow({
+    where: { id: session.user.id },
+    select: {
+      premium: {
+        select: {
+          id: true,
+          tier: true,
```
elie222, line 102: We don't need `id` or `tier` here, right?

apps/web/utils/actions.ts
```diff
@@ -109,7 +125,10 @@ export async function deleteAccountAction() {
     captureException(error);
   }
 
-  await prisma.user.delete({ where: { email: session.user.email } });
+  if (lemonSqueezySubscriptionId)
+    await cancelSubScriptionForUser(lemonSqueezySubscriptionId);
+
+  // await prisma.user.delete({ where: { email: session.user.email } });
```
elie222, line 131: Uncomment. We do want to delete the user.

(We do want to keep the user in Lemon though)

elie222 (changes requested):
Left 2 comments

PR #171, opened 2024-03-20
Update README.md

axelsimon:
Update README to be more explicit about being a Gmail-specific application.


elie222:
Has some mistakes. But thank you. Will add a note.

axelsimon:
Ah yes, good catch, some mishap with my ctrl+F'ing. Thanks though!

PR #174, opened 2024-03-22
prevent signin button to popup while switching to other page

### LINKED ISSUE #168: Bug: Signin button pops when switching to other page
If I click on the usage button here which will come when you click on the profile section, while it is getting reload I can see the "Signin" button at the same place.

image

Deeptanshu-sankhwar:
This PR fixes #168 

It conditionally checks the state of the react hook useSession(), and renders the button strategically based on the states of authenticated, loading and unathenticated


Deeptanshu-sankhwar:
This PR has been updated @elie222. Kindly review it.

PR #175, opened 2024-03-24
Disappear email when unsubscribe is clicked in simple mode

### LINKED ISSUE #155: Simple mode: email should disappear after you click unsubscribe

Deeptanshu-sankhwar:
This PR fixes #155 

It adds a new state for messages to display on the simple mode and adds a new handler module to pop a message when its unsubscribed.

### DIFF
--- a/apps/web/app/(app)/simple/SimpleList.tsx
+++ b/apps/web/app/(app)/simple/SimpleList.tsx
@@ -120,7 +120,7 @@ export function SimpleList(props: {
             });
           }}
         >
-          {toArchive && toArchive.length ? "Archive and Continue" : "Continue"}
+          {toArchive.length ? "Archive and Continue" : "Continue"}
         </HoverButton>
       </div>
 
--- a/apps/web/app/(app)/simple/SimpleList.tsx
+++ b/apps/web/app/(app)/simple/SimpleList.tsx
@@ -55,7 +55,7 @@ export function SimpleList(props: {
   const { toHandleLater, onSetHandled, onSetToHandleLater } =
     useSimpleProgress();
 
-  const [unsubscribed, setUnsubscribed] = useState([]);
+  const [unsubscribed, setUnsubscribed] = useState(new Set());
   const router = useRouter();
 
   const [parent] = useAutoAnimate();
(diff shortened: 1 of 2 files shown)


apps/web/app/(app)/simple/SimpleList.tsx
```diff
@@ -47,28 +47,33 @@ import { HtmlEmail } from "@/components/email-list/EmailPanel";
...
 
+  const [messages, setMessages] = useState(props.initialMessages);
   const router = useRouter();
 
   const [parent] = useAutoAnimate();
 
   const [isPending, startTransition] = useTransition();
 
-  const toArchive = props.messages
+  const toArchive = messages
```
elie222: We still want to archive unsubscribed messages. This would skip that.

This should work better. But don't use this filter in `toArchive`.

```ts
const [unsubscribed, ...] = useState()
```

apps/web/app/(app)/simple/SimpleList.tsx
```diff
@@ -47,28 +47,33 @@ import { HtmlEmail } from "@/components/email-list/EmailPanel";
...
+  const handleUnsubscribe = (id) => {
+    setMessages(messages.filter((message) => message.id !== id));
+  };
+
   return (
     <>
       <div className="mt-8 grid gap-4" ref={parent}>
-        {props.messages
+        {messages
           .filter((m) => !toHandleLater[m.id])
```
elie222: And then here you can filter by unsubbed `!toHandleLater[m.id] && !unsubscribed[m.id]`

apps/web/app/(app)/simple/SimpleList.tsx
```diff
@@ -78,12 +93,15 @@ export function SimpleList(props: {
...
             );
           })}
-        {props.messages.length === 0 && (
-          <Celebration message="All emails handled!" />
-        )}
+        {props.messages.filter(
+          (m) =>
+            !toHandleLater[m.id] &&
+            !unsubscribed.find((unsub) => unsub.id === m.id),
+        ).length === 0 && <Celebration message="All emails handled!" />}
```
elie222: Looks like this is a copy paste of the above. Can do `const filteredMessages = ...` instead and then use this in each spot.

apps/web/app/(app)/simple/SimpleList.tsx
```diff
@@ -65,11 +66,25 @@ export function SimpleList(props: {
...
 
+  const handleUnsubscribe = (id) => {
+    const message = props.messages.find((item) => item.id === id);
+    if (message) {
+      setUnsubscribed((currentUnsubscribed) => [
+        ...currentUnsubscribed,
+        message,
+      ]);
+    }
+  };
```
elie222: Seems overly complicated. Why not just add the id to the array? Why the need for find?

Also, you could make it a Set or object instead of an array of ids. A bit more efficient and then wouldn't need to do `.find`.

elie222:
Thanks! Much cleaner now!

PR #178, opened 2024-04-17
feat: Add double-click functionality to open newsletter rows

foobarnes:
I need a way to quickly open/glimpse the details of a Newsletter row. I need to see more than just the sender to identify whether a newsletter is one I need to unsubscribe from or not.


elie222:
Very cool!

PR #198, opened 2024-07-08
Feat: Dockerized for frontend

0xrushi:
Removes the need to set up node and pnpm and next etc.

To run:

set the `.env` file and

```
docker compose up -d
```

For docker desktop, you might need to enable this host networking

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,5 +47,4 @@ dist
 .react-email
 
 .sentryclirc
-.vscode/settings.json
 .pnpm-store/*
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,5 @@ dist
 .react-email
 
 .sentryclirc
+.vscode/settings.json
+.pnpm-store/*
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "next dev -p 3000 -H 0.0.0.0",
     "build": "NODE_OPTIONS=--max_old_space_size=8192 prisma migrate deploy && next build",
     "start": "next start",
     "lint": "next lint",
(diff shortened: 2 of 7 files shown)


0xrushi:
main branch isn't working for me that is another issue, i'll post it under issues later. but it works right now with dev.

.gitignore
```diff
@@ -47,3 +47,5 @@ dist
 .react-email
 
 .sentryclirc
+.vscode/settings.json
```
elie222: Hey, keep `.vscode`. Helpful for people to share common settings.

Other repos keep it too. eg.: https://github.com/calcom/cal.com

Can share recommended extensions with it for example:
https://github.com/calcom/cal.com/blob/main/.vscode/extensions.json

apps/web/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "next dev -p 3000 -H 0.0.0.0",
```
elie222: Curious what the reason for this change was? Also that by default it opens at port 3000 anyway.

package.json
```diff
@@ -11,6 +11,7 @@
     "ncu": "ncu -u -ws"
   },
   "devDependencies": {
+    "prisma": "^4.0.0",
```
elie222: You shouldn't need to add this here. It's add in the `web` app. Also the version is wrong. We're already on prisma 5.

apps/web/package.json
0xrushi: removed the port, 0.0.0.0 is needed to access the frontend outside of docker

elie222:
Thanks!

elie222:
Hi @0xrushi, I merged this into `main` just now: https://github.com/elie222/inbox-zero/pull/199
`dev` wasn't really being used.

I had to remove args from the turborepo file as that wasn't an allowed option and broke the build. Was that field necessary?

0xrushi:
Oh ok, not sure, I didn't use turbo but pnpm dev to build it. 
But I read there was the turbo way as well.

PR #202, opened 2024-07-19
Added active class to alpha mail sidebar

Vishvsalvi:
image

The active class remains on the sidebar

## Summary by CodeRabbit

- **New Features**
	- Enhanced sidebar navigation to incorporate search parameters for dynamic link generation.
	- Improved logic for active link styling, allowing for more accurate indication of selected items.

- **Bug Fixes**
	- Refined active state determination for navigation links to support both exact and partial matches, enhancing usability.

### DIFF
--- a/apps/web/components/SideNav.tsx
+++ b/apps/web/components/SideNav.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { Fragment } from "react";
-import { usePathname } from "next/navigation";
+import { usePathname, useSearchParams } from "next/navigation";
 import Link from "next/link";
 import {
   Dialog,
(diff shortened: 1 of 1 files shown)


apps/web/components/SideNav.tsx
```diff
@@ -392,14 +398,15 @@ function Links(props: { path: string; links: NavItem[] }) {
 
 function NavLink(props: { path: string; link: NavItem }) {
   const { link } = props;
+  const isActive = props.path === link.href || props.path.includes(link.href);
```
elie222, line 401: What was the reason for `includes` here?
But also `includes` would match `===` too.

elie222 (changes requested):
One question

apps/web/components/SideNav.tsx
Vishvsalvi: The `includes` matches the query string, `===` doesn't match the query string it is for the paths.

elie222: If `props.path.includes(link.href)` is true then it must be that `props.path === link.href` is true

Vishvsalvi: Ohh I haven't checked that yet! If that's the case I'll make the changes

elie222:
We changed to shadcn sidebar since this PR was opened

PR #211, opened 2024-08-13
Fixes #115 : Pagination is added via load more button

### LINKED ISSUE #115: Mail: Pagination
Add pagination to mail.

madhav-relish:
### Description

This PR fixes the pagination issue in mail

### Changes

- When the user clicks on load more button, the api call is made
- While the api is fetching the data a loader is being displayed
- User can perform any action and the refetch triggers to make sure data gets updated
- After deleting or performing any action, the page doesn't resets to first page, the user stays where they were before

Note: Is there a better way of doing this? any suggestions?

https://github.com/user-attachments/assets/cea53da8-42fd-4cb3-bd9b-e236bb8690d0

Related Issue

Fixes #115 

## Summary by CodeRabbit

- **New Features**
	- Introduced a "Load More" feature for email threads, allowing users to dynamically load additional emails.
	- Enhanced user interface with a loading spinner on the "Load More" button for better user feedback during the loading process.
	- Implemented infinite scrolling for email threads, improving navigation through large email lists.

- **Bug Fixes**
	- Improved state management to ensure the latest email threads are displayed correctly based on user interactions.


apps/web/app/(app)/mail/page.tsx
```diff
@@ -25,9 +25,15 @@ export default function Mail({
...
       keepPreviousData: true,
       dedupingInterval: 1_000,
+      revalidateOnFocus: false,
     },
   );
 
+  // States to handle more emails for the load more feature
+  const [allThreads, setAllThreads] = useState(data?.threads || []);
+  const [nextPageTokenId, setNextPageTokenId] = useState(data?.nextPageToken);
+  const [isLoadMoreLoading, setIsLoadMoreLoading] = useState(false);
```
elie222: Can use `useSWRInfinite` for potentially cleaner solution here:
https://swr.vercel.app/docs/pagination#example-1-index-based-paginated-api

apps/web/app/(app)/mail/page.tsx
```diff
@@ -68,17 +74,50 @@ export default function Mail({
...
+        const newData: ThreadsResponse = await res.json();
+
+        // update the threads and nextPageToken
+        return {
+          threads: [...(currentData?.threads || []), ...newData.threads],
+          nextPageToken: newData.nextPageToken,
+        };
+      },
+      { revalidate: false },
+    ).finally(() => setIsLoadMoreLoading(false));
```
elie222: Not critical but use try/await/catch instead of finally to be more consistent with the rest of the codebase.

apps/web/components/email-list/EmailList.tsx
```diff
@@ -33,13 +33,24 @@ import {
...
 import { selectedEmailAtom } from "@/store/email";
 import { categorizeAction } from "@/utils/actions/categorize";
+import { Button, ButtonLoader } from "@/components/ui/button";
+import { ChevronDown } from "lucide-react";
 
 export function List(props: {
   emails: Thread[];
   type?: string;
   refetch: (removedThreadIds?: string[]) => void;
+  isLoadMore?: boolean;
```
elie222: Could name better. eg `showLoadMore`. AI is good at helping with better naming.

apps/web/components/email-list/EmailList.tsx
```diff
@@ -33,13 +33,24 @@ import {
...
+  isLoadMoreLoading?: boolean;
```
elie222: `isLoadingMore` would be simpler

apps/web/components/email-list/EmailList.tsx
```diff
@@ -33,13 +33,24 @@ import {
...
+  handleLoadMoreThreads?: () => void;
```
elie222: `handleLoadMore` probably fine here

apps/web/components/email-list/EmailList.tsx
```diff
@@ -472,6 +492,25 @@ export function EmailList(props: {
...
+                  <Button
+                    variant="outline"
+                    className="w-full py-4"
+                    onClick={handleLoadMoreThreads}
+                  >
+                    {isLoadMoreLoading ? (
+                      <ButtonLoader />
+                    ) : (
+                      <>
+                        {" "}
```
elie222: Why the space here?
From when I ran it the UI didn't look great here.

To be more inline with the rest of the UI, you could have a spinner replace the icon. And have the icon to the left.

See the `show more` button on the Bulk Unsubscribe page for example:
CleanShot 2024-08-29 at 09 32 30@2x

elie222 (changes requested):

apps/web/app/(app)/mail/page.tsx
madhav-relish: Done

apps/web/app/(app)/mail/page.tsx
madhav-relish: No need of this anymore!

apps/web/components/email-list/EmailList.tsx
madhav-relish: Done

apps/web/components/email-list/EmailList.tsx
madhav-relish: Done

apps/web/components/email-list/EmailList.tsx
madhav-relish: Done

apps/web/components/email-list/EmailList.tsx
madhav-relish: Done!

The "ButtonLoader" component has some css issue, it wobbles at it's axis. Saw it at a few other places in the UI. 
One Simple fix was to change the size to 5, but then it would have caused UI inconsistency so I didn't touch it!
Just letting you know!

apps/web/app/(app)/mail/page.tsx
```diff
@@ -20,65 +20,80 @@ export default function Mail({
...
+    pageIndex: number,
+    previousPageData: ThreadsResponse | null,
+  ) => {
+    if (previousPageData && !previousPageData.nextPageToken) return null;
+    // For the first page, use the base query
+    if (pageIndex === 0)
+      return `/api/google/threads?${new URLSearchParams(query as any).toString()}`;
+    // For subsequent pages, append the nextPageToken
+    return `/api/google/threads?nextPageToken=${previousPageData?.nextPageToken}&${new URLSearchParams(query as any).toString()}`;
+  };
```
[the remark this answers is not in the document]
madhav-relish: Done!

apps/web/app/(app)/mail/page.tsx
```diff
@@ -20,65 +20,80 @@ export default function Mail({
...
+
+          return currentData.map((page) => ({
+            ...page,
+            threads: page.threads.filter(
               (t) => !removedThreadIds.includes(t.id),
-            ) || [];
-          return { threads, nextPageToken: undefined };
+            ),
+          }));
         },
```
elie222, line 64: Hey, I realise this was messy code before. But what's the reason the optimistic update was removed?

apps/web/app/(app)/mail/page.tsx
```diff
@@ -20,65 +20,80 @@ export default function Mail({
...
 
-  // store `refetch` in the atom so we can refresh the list upon archive via command k
-  // TODO is this the best way to do this?
+  // Set up the refetch function in the atom store
   const setRefetchEmailList = useSetAtom(refetchEmailListAtom);
   useEffect(() => {
     setRefetchEmailList({ refetch });
   }, [refetch, setRefetchEmailList]);
 
+  const handleLoadMore = () => {
```
elie222: Minor, but you can do `useCallback` here.

apps/web/components/email-list/EmailList.tsx
```diff
@@ -472,6 +488,28 @@ export function EmailList(props: {
                   />
                 );
               })}
+              {/* Load more button to fetch more data */}
```
elie222: can drop this comment

apps/web/components/email-list/EmailList.tsx
```diff
@@ -472,6 +488,28 @@ export function EmailList(props: {
...
+              {showLoadMore && (
+                <div className="pb-2">
```
elie222: styling for this button is off in your video. adding px-4 may fix this. you could also drop the wrapper div, and just do mb-2 mx-4 on the button

apps/web/components/email-list/EmailList.tsx
```diff
@@ -472,6 +488,28 @@ export function EmailList(props: {
...
+                  <Button
+                    variant="outline"
+                    className="w-full py-4"
```
elie222: is py-4 needed? there's button size prop if you want to adjust sizing but not sure that's needed here

apps/web/app/(app)/mail/page.tsx
elie222: But also if it works from your video, then maybe it's fine.

apps/web/app/(app)/mail/page.tsx
```diff
@@ -20,65 +20,80 @@ export default function Mail({
...
+    pageIndex: number,
+    previousPageData: ThreadsResponse | null,
+  ) => {
+    if (previousPageData && !previousPageData.nextPageToken) return null;
+    // For the first page, use the base query
+    if (pageIndex === 0)
+      return `/api/google/threads?${new URLSearchParams(query as Record<string, string>).toString()}`;
+    // For subsequent pages, append the nextPageToken
+    return `/api/google/threads?nextPageToken=${previousPageData?.nextPageToken}&${new URLSearchParams(query as any).toString()}`;
+  };
```
elie222: cleaner way to do this is without the if/else around the whole row as most of it is duplicated.

something like this:

```ts
const getKey = (pageIndex: number, previousPageData: ThreadsResponse | null) => {
  if (previousPageData && !previousPageData.nextPageToken) return null;

  const queryParams = new URLSearchParams(query as Record<string, string>);
  
  if (previousPageData?.nextPageToken) {
    queryParams.set('nextPageToken', previousPageData.nextPageToken);
  }

  return `/api/google/threads?${queryParams.toString()}`;
};
```

elie222 (changes requested):

apps/web/app/(app)/mail/page.tsx
madhav-relish: Just for the simplicity of the logic and code.

apps/web/app/(app)/mail/page.tsx
madhav-relish: Done!

apps/web/components/email-list/EmailList.tsx
madhav-relish: Dropped.

apps/web/components/email-list/EmailList.tsx
madhav-relish: Removed the wrapper div and added mb-2 for the button, mx-4 was creating overflow

apps/web/components/email-list/EmailList.tsx
madhav-relish: Added size prop and removed py-4

apps/web/app/(app)/mail/page.tsx
madhav-relish: Done!

madhav-relish:
@elie222  Any more changes required here?

elie222:
Seems to be working great on the Mail page. Wondering if we can also get it to work on the view when a user views emails on other pages. eg `bulk unsubscribe` page doesn't show `load more`:
CleanShot 2024-10-17 at 23 09 03@2x

Probably need to extract this into its own hook:

```tsx
  const query: ThreadsQuery = searchParams.type
    ? { type: searchParams.type }
    : {};

  const getKey = (
    pageIndex: number,
    previousPageData: ThreadsResponse | null,
  ) => {
    if (previousPageData && !previousPageData.nextPageToken) return null;
    const queryParams = new URLSearchParams(query as Record<string, string>);
    // Append nextPageToken for subsequent pages
    if (pageIndex > 0 && previousPageData?.nextPageToken) {
      queryParams.set("nextPageToken", previousPageData.nextPageToken);
    }
    return `/api/google/threads?${queryParams.toString()}`;
  };

  const { data, size, setSize, isLoading, error, mutate } =
    useSWRInfinite<ThreadsResponse>(getKey, {
      keepPreviousData: true,
      dedupingInterval: 1_000,
      revalidateOnFocus: false,
    });

  const allThreads = data ? data.flatMap((page) => page.threads) : [];
  const isLoadingMore =
    isLoading || (size > 0 && data && typeof data[size - 1] === "undefined");
  const showLoadMore = data ? !!data[data.length - 1]?.nextPageToken : false;
  ```
 
  e.g. for usage in `apps/web/app/(app)/stats/NewsletterModal.tsx`

PR #254, opened 2024-11-01
Add GROQ support

0xrushi:
## Release Notes

- **New Features**
	- Added support for the new AI provider, GROQ, enhancing AI settings functionality.
	- Introduced new error handling functions for specific GROQ API errors.
	- Updated configuration to include GROQ in provider and model options.
	- Added optional environment variable `GROQ_API_KEY` for server configuration.
	- Expanded cost calculation for AI model usage, including new models.

- **Bug Fixes**
	- Improved error handling logic to provide more detailed error messages during settings updates.

- **Chores**
	- Added new dependency `@ai-sdk/groq` to the project for GROQ support.
	- Updated existing dependencies for improved performance and compatibility.

PR #255, opened 2024-11-01
Add OLLAMA support

0xrushi:
This is similar to the PR for [GROQ support](https://github.com/elie222/inbox-zero/pull/254) which I know works, I'm keeping it open till someone with ollama can run llama 70b and test it.


apps/web/utils/error.ts
```diff
@@ -101,6 +101,35 @@ export function isOpenAIRetryError(error: unknown): boolean {
...
+  );
+}
+
+function isKnownOllamaError(error: APICallError): boolean {
+  return (
+    isOllamaInvalidApiKeyError(error) ||
+    isOllamaRateLimitError(error) ||
+    isOllamaQuotaExceededError(error)
+  );
+}
```
elie222, line 131: We can remove this if this is just guesswork

apps/web/utils/llms/config.ts
```diff
@@ -1,18 +1,21 @@
...
   ANTHROPIC: "anthropic",
+  OLLAMA: "ollama",
 };
 
 export const Model = {
   GPT_4O: "gpt-4o",
   GPT_4O_MINI: "gpt-4o-mini",
   CLAUDE_3_5_SONNET_BEDROCK: "anthropic.claude-3-5-sonnet-20241022-v2:0",
   CLAUDE_3_5_SONNET_ANTHROPIC: "claude-3-5-sonnet-20241022",
+  OLLAMA_MODEL: "llama3.1:70b",
```
elie222, line 12: Ollama doesn't have one model. You can use many models with Ollama

apps/web/utils/llms/index.ts
```diff
@@ -65,6 +66,17 @@ function getModel({ aiProvider, aiModel, aiApiKey }: UserAIFields) {
     };
   }
 
+  if (provider === Provider.OLLAMA) {
+    const model = aiModel || Model.OLLAMA_MODEL;
+    return {
+      provider: Provider.OLLAMA,
+      model,
+      llmModel: createOllama({
+        baseURL: "https://api.ollama.com",
```
elie222, line 75: Would be better to set this as an env var so that people that self-host could even point this to their local Ollama if they want.

elie222 (changes requested):
Made some comments

apps/web/utils/llms/index.ts
elie222: Also what even is https://api.ollama.com/? It doesn't exist

0xrushi: Yes, that's right! It’s from the Ollama documentation. I figured only tech-savvy users would be using Ollama, so they’d know that this would be replaced.

Regarding the comment above, claude also has multiple models. how should we handle this adn the API url? should we mention it in the README?

elie222: Great way to confuse tech savy users 😂
Env vars are used for fields that need to be replaced

elie222:
Added Ollama support in a different PR

PR #355, opened 2025-02-23
Add mail labels

RicSala:
Add coloured labels to the `EmailListItem` component in the `apps/web` directory.:

image

* [`apps/web/components/email-list/EmailListItem.tsx`](diffhunk://#diff-a6c1b93e9ad1ebf6d08361bc4c798d3c44b0a19965e50c9f3d7b47d5bc652686R24-R30): Imported `useLabels`, `UserLabel`, `Badge`, `MoreVertical`, `HoverCard`, and `isDefined` to support the new label functionality.
* [`apps/web/components/email-list/EmailListItem.tsx`](diffhunk://#diff-a6c1b93e9ad1ebf6d08361bc4c798d3c44b0a19965e50c9f3d7b47d5bc652686R81-R96): Added logic to fetch and process user labels using the `useLabels` hook, and computed label-related states (`threadLabels`, `hasLabels`, `hasMoreLabels`).
* [`apps/web/components/email-list/EmailListItem.tsx`](diffhunk://#diff-a6c1b93e9ad1ebf6d08361bc4c798d3c44b0a19965e50c9f3d7b47d5bc652686R163-R188): Integrated the `LabelsGroup` component to display labels within the email list items and added a hover card for additional labels.

New component for label rendering:

* [`apps/web/components/email-list/EmailListItem.tsx`](diffhunk://#diff-a6c1b93e9ad1ebf6d08361bc4c798d3c44b0a19965e50c9f3d7b47d5bc652686R269-R295): Defined the `LabelsGroup` component to handle the display of user labels with optional wrapping and styling based on label colors.

Enhancements to user labels:

* [`apps/web/hooks/useLabels.ts`](diffhunk://#diff-a60925971bd29a8486c844d6bc0be46e98ec3487187b3444a95c200c6b8bfcf6L6-R14): Extended the `UserLabel` type to include color properties (`textColor` and `backgroundColor`).

## Summary by CodeRabbit

- **New Features**
  - Enhanced email thread display by showcasing user labels as colored badges.
  - Introduced a hoverable interface to reveal additional labels when there are more than can be immediately displayed.
  - Updated user label type to include optional color properties for improved visual representation.


elie222:
Thanks. Take a look at the nitpick comments from CodeRabbit. They'd be good to use here. @RicSala

apps/web/components/email-list/EmailListItem.tsx
```diff
@@ -137,6 +159,32 @@ export const EmailListItem = forwardRef(
...
+                            e.preventDefault();
+                          }}
+                        >
+                          <MoreVertical
+                            size={16}
+                            className="text-muted-foreground"
+                          />
+                        </Button>
+                      </HoverCard>
+                    )}
```
elie222: make this whole thing a component. i think all the new code you added can be a component

RicSala:
> Thanks. Take a look at the nitpick comments from CodeRabbit. They'd be good to use here. @RicSala

I was on it! They are fixed and gone now!

apps/web/hooks/useLabels.ts
```diff
@@ -3,7 +3,15 @@ import useSWR from "swr";
...
-export type UserLabel = { id: string; name: string; type: "user" };
+export type UserLabel = {
+  id: string;
+  name: string;
+  type: "user";
+  color?: {
+    textColor: string;
+    backgroundColor: string;
+  };
+};
```
elie222, line 14: would be better to just reuse the type from gmail, but okay as is. gmail a bit annoying as it makes all the fields optional. so at that point may just make sense to retype like we have here. i guess we'll just leave as is

apps/web/components/email-list/EmailListItem.tsx
elie222: can email move it into its own file in the email-list folder

elie222: possible we reuse this in other places in the app

apps/web/hooks/useLabels.ts
RicSala: so... wanna change this one? Not sure what path (either typing our own or using gmail types) you wanna take. I'm completely fine updating this!

apps/web/components/email-list/EmailListItem.tsx
RicSala: done 

btw, I used the badge from /ui folder because the other one (the one in components) was a bit rigid for the gmail color schemas. 

I could either:

1- leave it as it is (using /ui/badge)
2- make /components/badge more flexible (just a bit concerned about giving it "too much flexibility"), for example adding a style attr
3- Creating a "emailLabel" component, that uses the shadcn badge (ui/badge), and consistently use that one for email labels
4- Add the rest of the colors to badgeVariants

In my opinion, 1 and 3 are the cleanest, but lmk if you think otherwise.

apps/web/hooks/useLabels.ts
elie222: leave as is. dont worry

apps/web/components/email-list/EmailListItem.tsx
```diff
@@ -71,6 +76,21 @@ export const EmailListItem = forwardRef(
 
     const cta = findCtaLink(lastMessage.textHtml);
 
+    const { userLabels } = useLabels();
```
elie222, line 79: so to optimise here, don't do uselabels on every row item. but do it a level above, and pass in the labels. basically any logic here is being run on every single row. and you could just have it run once for the table

apps/web/components/email-list/EmailListItem.tsx
```diff
@@ -71,6 +76,21 @@ export const EmailListItem = forwardRef(
...
+
+    const threadLabels = useMemo(() => {
```
elie222, line 81: see the nitpick coderabbit notes:
https://github.com/elie222/inbox-zero/pull/355#pullrequestreview-2635583763
CleanShot 2025-02-23 at 16 26 29@2x

apps/web/components/email-list/EmailListItem.tsx
RicSala: @elie222 ☝🏻

apps/web/components/email-list/EmailListItem.tsx
RicSala: As we are using SWR, actually I think only one request will be made! Others will be dedup (afaik).

I can move it to the parent, just wanted to avoid cluttering more the attributes of EmailListItem comp!

apps/web/components/email-list/EmailListItem.tsx
RicSala: done!

apps/web/components/email-list/EmailListItem.tsx
elie222: Just asked Cursor. What it had to say. I considered your approach too, but I agree with the AI 😂

Cursor:


Right.Let me analyze this PR discussion about the `useLabels` hook in `EmailListItem`:

Both perspectives have valid points, but I would side with @elie222's suggestion for these reasons:

1. **Component Responsibility**
   - Even though SWR will deduplicate the actual API requests, the hook execution itself still runs for every row item
   - The `EmailListItem` component should ideally be focused on presentation, not data fetching
   - Moving data fetching up reduces the cognitive load when reading the item component

2. **Performance Optimization**
   - While SWR handles API deduplication, the hook still:
     - Creates new state instances per row
     - Runs its internal logic for each row
     - Triggers re-renders on each row when data changes
   - With many emails in the list, this overhead adds up

3. **Props vs. Hooks**
   - Props make dependencies explicit and easier to test
   - Props allow better control over when re-renders happen
   - The parent component can optimize when labels need to be refetched

Here's how I would refactor it:

```typescript:apps/web/components/email-list/EmailList.tsx
// ... existing code ...
const { userLabels } = useLabels();

return (
  <div>
    {threads.map(thread => (
      <EmailListItem 
        key={thread.id}
        userLabels={userLabels}
        thread={thread}
        // ... other props ...
      />
    ))}
  </div>
);
```

```typescript:apps/web/components/email-list/EmailListItem.tsx
// ... existing code ...
export const EmailListItem = forwardRef(
  (
    props: {
      userLabels: Label[];  // Add this prop
      userEmailAddress: string;
      thread: Thread;
      // ... other props ...
    },
    ref: ForwardedRef<HTMLLIElement>,
  ) => {
    // Remove useLabels() hook
    // ... rest of component ...
  }
);
```

While the extra prop might seem like clutter, it's a worthwhile tradeoff for better performance and clearer component responsibilities. The parent component is a more appropriate place for data fetching that affects multiple child components.

elie222: What I asked it btw:


what do you think about this discussion on a pr?


apps/web/components/email-list/EmailListItem.tsx
@@ -71,6 +76,21 @@ export const EmailListItem = forwardRef(

    const cta = findCtaLink(lastMessage.textHtml);

    const { userLabels } = useLabels();
Owner
@elie222 elie222 yesterday
so to optimise here, don't do uselabels on every row item. but do it a level above, and pass in the labels. basically any logic here is being run on every single row. and you could just have it run once for the table

Contributor
Author
@RicSala RicSala yesterday
As we are using SWR, actually I think only one request will be made! Others will be dedup (afaik).

I can move it to the parent, just wanted to avoid cluttering more the attributes of EmailListItem comp!

elie222:
We have a Gmail provider that I used for other pages just now:
https://github.com/elie222/inbox-zero/pull/367

Can probably do the same thing for Mail page. The gmail provider just stores the state once for us so a little more efficient although probably minor difference.

elie222:
Thanks! This was merged in another PR!

Issue #345, opened 2025-02-17
Mail: auto advance


elie222:
When you handle an email (E.g. archiving it), you should advance automatically to the next email in the list.

gshiva53:
Opened a pull request #369.

elie222:
Completed by https://github.com/elie222/inbox-zero/pull/369

Thanks!

PR #369, opened 2025-03-05
Auto Advance Email on Archive

gshiva53:
Hi, 

I am a first time contributor and I worked on the issue #345 Auto advance mail. The functionality works for archive for now but can work with other actions if needed. 

Related to #345. 

Demo Video: [https://www.loom.com/share/a2b58b50466b49169d3d6fa83e6bbbd2?sid=2e0e800b-71cb-42b0-8601-63134f6cbd3d](https://www.loom.com/share/a2b58b50466b49169d3d6fa83e6bbbd2?sid=2e0e800b-71cb-42b0-8601-63134f6cbd3d)

## Summary by CodeRabbit

- **New Features**
  - Archiving an email now automatically shifts focus to the next or previous email thread for smoother continuity.

- **Bug Fixes**
  - Adjusted the behavior after archiving so that the email panel remains open, enhancing user control throughout the process.

- **Changes**
  - Removed the `rejectPlan` function from the `EmailListItem` component's props, altering its interaction with parent components.


apps/web/components/email-list/EmailList.tsx
```diff
@@ -188,6 +188,27 @@ export function EmailList({
     [openThreadId, threads],
   );
 
+  const prevOrNextRowId = useMemo(() => {
```
elie222: i'd pull this out of the function, and then just call it where needed:
`setOpenThreadId(prevOrNextRowId);`
as we don't need this anywhere else. it's really just a helper function.

apps/web/components/email-list/EmailList.tsx
```diff
@@ -216,6 +237,8 @@ export function EmailList({
 
   const onArchive = useCallback(
     (thread: Thread) => {
+      setOpenThreadId(prevOrNextRowId);
```
elie222: i'd probably leave onArchive as is, and then just put this logic where we had `close()`.

also, if this is the only thread, in the list we could just call close() as before

elie222 (changes requested):
thanks! left 2 comments!

elie222:
Great that you added a video too! Super helpful. Also see the cla task that's pending

apps/web/components/email-list/EmailList.tsx
gshiva53: Hi, thanks for the comments. 

I have created a function in the latest commit.

apps/web/components/email-list/EmailList.tsx
gshiva53: yeah, makes more sense to have the archive function only have that functionality. 

I have moved the logic to where we had `close()`.

apps/web/components/email-list/EmailPanel.tsx
```diff
@@ -57,7 +58,7 @@ export function EmailPanel({
             onPlanAiAction={() => onPlanAiAction(row)}
             onArchive={() => {
               onArchive(row);
-              close();
+              advanceToAdjacentThread();
```
gshiva53, line 61: Moved the logic outside of `onArchive()` and where `close()` was before as per comments.

apps/web/components/email-list/EmailList.tsx
```diff
@@ -270,6 +270,23 @@ export function EmailList({
     }, 100);
   }
 
+  function advanceToAdjacentThread() {
```
gshiva53, line 273: Moved the logic into a function which will only be called `onArchive()` instead of every render (possibly) as per comments.

apps/web/components/email-list/EmailList.tsx
```diff
@@ -270,6 +270,23 @@ export function EmailList({
...
+    const openedRowIndex = threads.findIndex(
+      (thread) => thread.id === openThreadId,
+    );
+
+    if (openedRowIndex === -1 || threads.length === 0 || threads.length === 1)
```
gshiva53: Close the email panel when there is only one thread because that thread would have been archived.

gshiva53:
Signed the CLA.

elie222:
Thanks!

PR #391, opened 2025-04-06
Made the email appear on fullscreen

### LINKED ISSUE #344: Mail: Full screen mode for emails
ATM, we open in split screen mode, but most people prefer to view their emails in full screen. We should add support for this. It also relates to https://github.com/elie222/inbox-zero/issues/342

rohanshrma222:
This PR makes the email appear on full screen,Removed the split screen mode.

Fixes #344

Result Video attached :

https://github.com/user-attachments/assets/8e9e1307-8364-466b-8826-60b113670da3

### DIFF
--- a/apps/web/components/email-list/EmailPanel.tsx
+++ b/apps/web/components/email-list/EmailPanel.tsx
@@ -1,4 +1,4 @@
-import { XIcon } from "lucide-react";
+import { XIcon, ArrowLeftIcon } from "lucide-react";
 import { ActionButtons } from "@/components/ActionButtons";
 import { Tooltip } from "@/components/Tooltip";
 import type { Thread } from "@/components/email-list/types";
@@ -6,6 +6,7 @@ import { Button } from "@/components/ui/button";
 import { PlanExplanation } from "@/components/email-list/PlanExplanation";
 import { useIsInAiQueue } from "@/store/ai-queue";
 import { EmailThread } from "@/components/email-list/EmailThread";
+import { usePathname } from "next/navigation";
 
 export function EmailPanel({
   row,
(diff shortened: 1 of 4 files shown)


apps/web/components/email-list/EmailPanel.tsx
```diff
@@ -31,6 +32,8 @@ export function EmailPanel({
   refetch: () => void;
 }) {
   const isPlanning = useIsInAiQueue(row.id);
+  const pathname = usePathname();
+  const isFullscreen = pathname.split("/").length > 3; // Checking if we're in /mail/type/threadId
```
elie222, line 36: super hacky. better to pass in props

elie222: you can't rely on the url having 3+ parts to decide how to show the component

elie222:
Hey, would be best if we did this like Gmail and kept the split screen option as it's already working. But also offered full screen mode.

Also I notice the path for the page is /mail/inbox/messageid.
no reason for it to be under /inbox. although gmail does do something like this with their urls doesn't seem a smart way to handle it.

may also be worth considering keeping the page as is and the full screen is more a modal type thing. similar to split screen mode, but it just covers the whole screen instead. this way you don't lose page context / need reload when you go back to inbox

rohanshrma222:
Thanks for the suggestion.
1. Like in Gmail they have their setting for fullscreen view and split-screen, so how do you want me to do it  add a button or 
   something?
2. Do you want to avoid a new route and make a fullscreen model there only?

slothmc-ctrl:
> Thanks for the suggestion.
> 
>     1. Like in Gmail they have their setting for fullscreen view and split-screen, so how do you want me to do it  add a button or
>        something?
> 
>     2. Do you want to avoid a new route and make a fullscreen model there only?

Why not both ? Modal and new page button ?

PR #430, opened 2025-05-10
Fix: Duplicate React Keys in Bulk Unsubscribe Component and Hydration Errors

### LINKED ISSUE #429: Duplicate React Keys in Bulk Unsubscribe Component and Hydration Errors
# Multiple UI Rendering Issues: Key Duplication and Hydration Errors

## 🛠️ Issue

The application is experiencing multiple rendering issues:

1. **React key duplication error** in the **Bulk Unsubscribe** component. React requires each item in a list to have a **unique `key` prop** to optimize DOM updates.

2. **HTML hydration errors** in two separate components:
   - Nested `<p>` tags in the Reply Zero feature
   - Invalid table structure in the Cold Email Blocker

> **Error Messages:**
> ```
> Encountered two children with the same key, `noreply@google.com`. Keys should be unique so that components maintain their identity across updates.
> ```
> 
> ```
> In HTML, <p> cannot be a descendant of <p>. This will cause a hydration error.
> ```
> 
> ```
> In HTML, <tr> cannot be a child of <div>. This will cause a hydration error.
> ```

#### 1. Duplicate React Keys
In the `BulkUnsubscribeSection` component, the `key` was generated using only the newsletter item's `name` (e.g., `noreply@google.com`). This led to **duplicate keys** when multiple newsletter items shared the same sender address.

Image

#### 2. Nested Paragraph Tags
In the `EnableReplyTracker` component, a `<p>` tag was wrapping a `<SectionDescription>` component, which itself renders as a `<p>` element. HTML standards don't allow nesting paragraph elements, causing hydration mismatches between server and client rendering.

#### 3. Invalid Table Structure
The `TestRulesContent` component in the Cold Email Blocker was rendering `<tr>` (table row) elements directly inside a `<div>` instead of inside proper table elements like `<table>`, `<tbody>`, etc. This violates HTML structure rules.

Image

Swastik19Nit:
#### 1. Robust Key Generation
To resolve the key duplication issue:

```diff
- const unsortedTableRows = rows?.map((item) => {
+ const unsortedTableRows = rows?.map((item, index) => {
    // ...
-   const row = <RowComponent key={item.name} ... />
+   const uniqueKey = `${item.name}-${item.value}-${item.inboxEmails}-${item.readEmails}-${index}`;
+   const row = <RowComponent key={uniqueKey} ... />
```

#### 2. Fixed Nested Paragraph Structure
Removed the outer paragraph element that was causing invalid nesting:

```diff
extraDescription={
  <div className="mt-4 text-left">
-   <p>
    <SectionDescription>We label your emails with:</SectionDescription>
-   </p>
    <SectionDescription>
      <Badge color="green">{NEEDS_REPLY_LABEL_NAME}</Badge> - emails you
      need to reply to.
    </SectionDescription>
    // ...
  </div>
}
```

#### 3. Proper Table Structure
Replaced div with appropriate table elements:

```diff
<LoadingContent loading={isLoading} error={error}>
  {data && (
-   <div>
-     {data.messages.map((message) => {
-       return (
+   <Table>
+     <TableBody>
+       {data.messages.map((message) => (
          <TestRulesContentRow
            key={message.id}
            message={message}
            userEmail={userEmail}
          />
-       );
-     })}
-   </div>
+       ))}
+     </TableBody>
+   </Table>
  )}
</LoadingContent>
```

## Summary by CodeRabbit

- **Refactor**
  - Improved uniqueness of row keys in the bulk unsubscribe section for more reliable rendering.
  - Updated message display in the cold email blocker test rules to use a semantic table structure for better accessibility and markup consistency.
  - Simplified the layout in the reply tracker enablement section by removing unnecessary wrapper elements.

### DIFF
--- a/apps/web/app/(app)/[emailAccountId]/bulk-unsubscribe/BulkUnsubscribeSection.tsx
+++ b/apps/web/app/(app)/[emailAccountId]/bulk-unsubscribe/BulkUnsubscribeSection.tsx
@@ -135,14 +135,16 @@ export function BulkUnsubscribeSection({
   const { selected, isAllSelected, onToggleSelect, onToggleSelectAll } =
     useToggleSelect(rows?.map((item) => ({ id: item.name })) || []);
 
-  const unsortedTableRows = rows?.map((item) => {
+  const unsortedTableRows = rows?.map((item, index) => {
     const readPercentage = (item.readEmails / item.value) * 100;
     const archivedEmails = item.value - item.inboxEmails;
     const archivedPercentage = (archivedEmails / item.value) * 100;
 
+    const uniqueKey = `${item.name}-${item.value}-${item.inboxEmails}-${item.readEmails}-${index}`;
+
     const row = (
       <RowComponent
-        key={item.name}
+        key={uniqueKey}
         item={item}
         userEmail={userEmail}
         emailAccountId={emailAccountId}
--- a/apps/web/app/(app)/[emailAccountId]/bulk-unsubscribe/BulkUnsubscribeSection.tsx
+++ b/apps/web/app/(app)/[emailAccountId]/bulk-unsubscribe/BulkUnsubscribeSection.tsx
@@ -135,16 +135,14 @@ export function BulkUnsubscribeSection({
   const { selected, isAllSelected, onToggleSelect, onToggleSelectAll } =
     useToggleSelect(rows?.map((item) => ({ id: item.name })) || []);
 
-  const unsortedTableRows = rows?.map((item, index) => {
+  const unsortedTableRows = rows?.map((item) => {
     const readPercentage = (item.readEmails / item.value) * 100;
     const archivedEmails = item.value - item.inboxEmails;
     const archivedPercentage = (archivedEmails / item.value) * 100;
 
-    const uniqueKey = `${item.name}-${item.value}-${item.inboxEmails}-${item.readEmails}-${index}`;
-
     const row = (
       <RowComponent
-        key={uniqueKey}
+        key={item.name}
         item={item}
         userEmail={userEmail}
         emailAccountId={emailAccountId}
--- a/apps/web/app/(app)/[emailAccountId]/cold-email-blocker/TestRules.tsx
+++ b/apps/web/app/(app)/[emailAccountId]/cold-email-blocker/TestRules.tsx
@@ -16,7 +16,7 @@ import { Separator } from "@/components/ui/separator";
 import { AlertBasic } from "@/components/Alert";
 import { EmailMessageCell } from "@/components/EmailMessageCell";
 import { SearchForm } from "@/components/SearchForm";
-import { TableCell, TableRow } from "@/components/ui/table";
+import { TableCell, TableRow, Table, TableBody } from "@/components/ui/table";
 import { CardContent } from "@/components/ui/card";
 import { testColdEmailAction } from "@/utils/actions/cold-email";
 import type { ColdEmailBlockerBody } from "@/utils/actions/cold-email.validation";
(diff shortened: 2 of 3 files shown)


apps/web/app/(app)/[emailAccountId]/bulk-unsubscribe/BulkUnsubscribeSection.tsx
```diff
@@ -135,14 +135,16 @@ export function BulkUnsubscribeSection({
   const { selected, isAllSelected, onToggleSelect, onToggleSelectAll } =
     useToggleSelect(rows?.map((item) => ({ id: item.name })) || []);
 
-  const unsortedTableRows = rows?.map((item) => {
+  const unsortedTableRows = rows?.map((item, index) => {
     const readPercentage = (item.readEmails / item.value) * 100;
     const archivedEmails = item.value - item.inboxEmails;
     const archivedPercentage = (archivedEmails / item.value) * 100;
 
+    const uniqueKey = `${item.name}-${item.value}-${item.inboxEmails}-${item.readEmails}-${index}`;
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/bulk-unsubscribe/BulkUnsubscribeSection.tsx]: might as well just use index if we're doing this, no?

also, why isn't item.name unique here? when does it happen that we have the same item.name twice?

elie222:
happy to accept the pr other than the index issue

apps/web/app/(app)/[emailAccountId]/bulk-unsubscribe/BulkUnsubscribeSection.tsx
Comment (Swastik19Nit) [reply apps/web/app/(app)/[emailAccountId]/bulk-unsubscribe/BulkUnsubscribeSection.tsx]: For the second part that why isn't item.name unique? When does it happen that we have duplicate names, the item.name represents the email sender (e.g., "noreply@google.com"). It's not unique because many services send different types of newsletters from the same email address and services like Google can send various types of emails (notifications, security alerts, promotions) all from "noreply@google.com".

For the first part we can use index but while using just index as the key would technically solve the duplicate key issue, it's not the best practice for React keys. In the official documentation it is written that "_You might be tempted to use an item’s index in the array as its key. In fact, that’s what React will use if you don’t specify a key at all. But the order in which you render items will change over time if an item is inserted, deleted, or if the array gets reordered. Index as a key often leads to subtle and confusing bugs_".

But we can change the key to just use the index if that's preferred. I will do that if you want to!!! @elie222 .

Swastik19Nit:
> happy to accept the pr other than the index issue

Sir if needed I can revert back to the original indexing code). But I guess it still needs some changes too. @elie222

Swastik19Nit:
> happy to accept the pr other than the index issue

@elie222 Hi Sir,
Thanks! I've undone the changes in the index!! Kindly look sir. Thanks

elie222:
Thanks!

elie222:
> > happy to accept the pr other than the index issue
> 
> Sir if needed I can revert back to the original indexing code). But I guess it still needs some changes too. @elie222

you can probably just do `key={index}`. im not sure there was any benefit to the complex key you had which anyway included index.

> For the second part that why isn't item.name unique? When does it happen that we have duplicate names, the item.name represents the email sender (e.g., "[noreply@google.com](mailto:noreply@google.com)"). It's not unique because many services send different types of newsletters from the same email address and services like Google can send various types of emails (notifications, security alerts, promotions) all from "[noreply@google.com](mailto:noreply@google.com)".

can you send a screenshot of what you're seeing. because these should be grouped on the bulk unsubscribe page. since it's the same sender (we used to care about sender name, but now it should just be based on email address)

Swastik19Nit:
> > > happy to accept the pr other than the index issue
> > 
> > 
> > Sir if needed I can revert back to the original indexing code). But I guess it still needs some changes too. @elie222
> 
> you can probably just do `key={index}`. im not sure there was any benefit to the complex key you had which anyway included index.
> 
> > For the second part that why isn't item.name unique? When does it happen that we have duplicate names, the item.name represents the email sender (e.g., "[noreply@google.com](mailto:noreply@google.com)"). It's not unique because many services send different types of newsletters from the same email address and services like Google can send various types of emails (notifications, security alerts, promotions) all from "[noreply@google.com](mailto:noreply@google.com)".
> 
> can you send a screenshot of what you're seeing. because these should be grouped on the bulk unsubscribe page. since it's the same sender (we used to care about sender name, but now it should just be based on email address)

I saw this error
Image
@elie222

PR #482, opened 2025-06-04
Fix Docker Image Build 

gowthamkishore3799:
Summary:

The Docker image pipeline isn’t updating version.txt as expected. This issue is due to the specific steps we use in the pipeline, so I’ve removed the step that pushes code and tags to avoid this problem.

## Summary by CodeRabbit

- **Chores**
	- Updated workflow to set a custom display name for test runs in GitHub Actions.
	- Adjusted Docker publishing workflow to stop pushing tags when updating version information.

### DIFF
--- a/.github/workflows/build_and_publish_docker.yml
+++ b/.github/workflows/build_and_publish_docker.yml
@@ -86,7 +86,6 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           branch: ${{ github.ref }}
           force_with_lease: true
-          tags: true
 
   build-docker:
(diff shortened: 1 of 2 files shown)


gowthamkishore3799:
Currently the version..txt isnt updated due to the github step we are using which should ideally push tag and along with file which isnt happening

So we are removing tags, which automatically updates the tag push along with file push

elie222:
So I accepted but ended up reverting. The issue is this:
https://github.com/elie222/inbox-zero/actions/runs/15375444526

The change made in the PR impacts a step after the error.

PR #493, opened 2025-06-06
Outlook auth

miha-yy:


apps/web/env.ts
```diff
@@ -12,6 +12,10 @@ export const env = createEnv({
     GOOGLE_CLIENT_SECRET: z.string().min(1),
     GOOGLE_ENCRYPT_SECRET: z.string(),
     GOOGLE_ENCRYPT_SALT: z.string(),
+    MICROSOFT_CLIENT_ID: z.string().min(1),
+    MICROSOFT_CLIENT_SECRET: z.string().min(1),
+    MICROSOFT_ENCRYPT_SECRET: z.string(),
+    MICROSOFT_ENCRYPT_SALT: z.string(),
```
elie222: We can make these optional so that people cna run the project without setting Microsoft keys

apps/web/utils/outlook/client.ts
```diff
@@ -0,0 +1,141 @@
...
+    throw error;
+  }
+};
+
+export const getAccessTokenFromClient = (client: Client): string => {
+  // This is a bit hacky but works - the client stores the token in its authProvider
+  const accessToken = (client as any)._authProvider._token;
+  if (!accessToken) throw new Error("No access token");
+  return accessToken;
+};
```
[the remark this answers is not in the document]
elie222: Is there a better way to do this? I know I did something hacky like this for Gmail, was this just copied over from that? I assume it'd be different for Outlook

elie222:
Also, please add some short notes on how to set it up / get the env keys. Even just 2 lines + a link is fine so I don't need to investigate when setting it up myself.

When the feature goes live for real we can add it to the README for future contributors / self-hosters.

apps/web/utils/actions/safe-action.ts
```diff
@@ -55,6 +55,9 @@ export const actionClient = baseClient
     // validate user owns this email
     const emailAccount = await prisma.emailAccount.findUnique({
       where: { id: emailAccountId },
+      include: {
```
elie222: Use `select` here. Fetch only the data we need. Also more secure, especially important when fetching account data.

apps/web/utils/auth.ts
```diff
@@ -4,19 +4,81 @@ import type { Prisma } from "@prisma/client";
...
+    },
+  },
+  "microsoft-entra-id": {
+    name: "microsoft",
+    tokenUrl: "https://login.microsoftonline.com/common/oauth2/v2.0/token",
+    getProfileData: async (accessToken: string) => {
+      const client = getOutlookContactsClient({ accessToken });
+      try {
+        const profileResponse = await client.getUserProfile();
+        console.log("profileResponse", profileResponse);
```
elie222: Remove console.log. Use logger.info/trace when we do want to log things.

apps/web/utils/auth.ts
```diff
@@ -404,13 +467,15 @@ export async function saveTokens({
...
   emailAccountId,
+  provider = "google",
 }: {
   tokens: {
     access_token?: string;
     refresh_token?: string;
     expires_at?: number;
   };
   accountRefreshToken: string | null;
+  provider?: string;
```
elie222: make this required instead?

apps/web/utils/auth.ts
```diff
@@ -404,13 +467,15 @@ export async function saveTokens({
   accountRefreshToken,
   providerAccountId,
   emailAccountId,
+  provider = "google",
```
elie222: don't default to google. should be explicitly passed in

apps/web/utils/outlook/client.ts
```diff
@@ -0,0 +1,164 @@
...
+  expiresAt?: number | null;
+};
+
+// Wrapper class to hold both the Microsoft Graph client and its access token
+export class OutlookClient {
+  private client: Client;
+  private accessToken: string;
+
+  constructor(accessToken: string) {
+    this.accessToken = accessToken;
```
elie222, line 24: note, we need to handle token refreshes. the access token will be fairly short lived. with google it has an auto refresh mechanism. i don't know if there's a similar lib for microsoft that does the same thing

apps/web/utils/actions/permissions.ts
```diff
@@ -12,7 +12,12 @@ const logger = createScopedLogger("actions/permissions");
 
 export const checkPermissionsAction = actionClient
   .metadata({ name: "checkPermissions" })
-  .action(async ({ ctx: { emailAccountId } }) => {
+  .action(async ({ ctx: { emailAccountId, provider } }) => {
```
elie222, line 15: btw `actionClient` likely needs changes. There's also middleware for GET requests in the middleware.ts file that probably needs adjusting too

PR #519, opened 2025-06-23
feat: Delayed actions

edulelis:
## Summary by CodeRabbit

* **New Features**
  * Introduced delayed actions for rules, allowing users to schedule actions (like archive, label, reply, etc.) to execute after a configurable delay (1 minute to 90 days).
  * Added UI controls to set action delays when creating or editing rules.
  * Included a new "Label + archive after a week" option during onboarding, when available.
  * Users can view and manage delayed actions in relevant rule and action interfaces.

* **Bug Fixes**
  * Improved tooltip positioning and consistency in UI components.

* **Documentation**
  * Added documentation describing the delayed actions feature, its use cases, and scheduling details.

* **Tests**
  * Added comprehensive tests for delayed action scheduling, execution, and cancellation.

* **Chores**
  * Updated validation and schema definitions to support delay configuration and new scheduling logic.


edulelis:
@coderabbitai review

apps/web/components/TooltipExplanation.tsx
```diff
@@ -19,18 +25,29 @@ interface TooltipExplanationProps
   extends React.HTMLAttributes<HTMLDivElement>,
     VariantProps<typeof tooltipIconVariants> {
   text: string;
+  side?: "top" | "right" | "bottom" | "left";
```
edulelis, line 30: Added this to better accommodate tooltips inside the modals.
The tooltips that are too close to the right, I aligned them to the left, and the ones that are too close the left I aligned them to the right. The default stays the same, which is "top"

apps/web/app/api/admin/scheduled-actions/[id]/cancel/route.ts
```diff
@@ -0,0 +1,58 @@
...
+    const session = await auth();
+
+    if (!session?.user?.email) {
+      return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
+    }
+
+    if (!isAdmin({ email: session.user.email })) {
+      return NextResponse.json({ error: "Forbidden" }, { status: 403 });
+    }
+
```
Comment (edulelis) [inline apps/web/app/api/admin/scheduled-actions/[id]/cancel/route.ts]: Extra endpoints to perform actions in the admin.
Explicitly checking for admin permissions.

apps/web/utils/ai/choose-rule/run-rules.ts
```diff
@@ -95,6 +99,24 @@ async function executeMatchedRule(
     gmail,
   });
 
+  // Cancel any existing scheduled actions for this message
+  if (!isTest) {
+    await cancelScheduledActions({
```
edulelis: If a delayed action should be re-run. This is going to cancel the old one.
Time 0: Email arrives
Time 1: Rule A executes → Schedule "Archive in 2 days"
Later on: User updates Rule A to "Archive in 1 day" → Rule A re-executes
        → Cancel previous "Archive in 2 days" 
        → Schedule new "Label and Archive in 1 day"
Cron: Only the new scheduled action will execute

apps/web/utils/delayed-actions.ts
```diff
@@ -0,0 +1,16 @@
+import { ActionType } from "@prisma/client";
+
+// Action types that do NOT support delayed execution
+const UNSUPPORTED_DELAYED_ACTIONS: ActionType[] = [
```
edulelis: This pattern is very permissive and will allow any new actions to have a delay configured by default.

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -857,7 +859,7 @@ export function RuleForm({
               variant="secondary"
               size="sm"
               onClick={() => {
-                append({ type: ActionType.LABEL });
+                append({ type: ActionType.LABEL, delayInMinutes: null } as any);
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: ideally without as any

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -867,9 +869,10 @@ export function RuleForm({
           </div>
         )}
 
-        <div className="mt-8 flex items-center justify-end space-x-2">
+        <div className="mt-[67px] flex items-center justify-end space-x-2">
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: reason for 67px hard coded?

apps/web/utils/delayed-actions.ts
elie222: Change to "supported"

apps/web/utils/reply-tracker/enable.ts
```diff
@@ -110,8 +110,11 @@ export async function enableReplyTracker({
               bcc: null,
               webhookUrl: null,
             },
+            delayInMinutes: null,
           },
-          ...(addDigest ? [{ type: ActionType.DIGEST }] : []),
+          ...(addDigest
+            ? [{ type: ActionType.DIGEST, delayInMinutes: null }]
```
elie222: do we really need to set null here? all the other fields aren't set to null explicitly, so is this needed?

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
Comment (edulelis) [reply apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: This is leftover from experimentation trying to line the action name and label.
From this:
Screenshot 2025-06-30 at 07 02 32
To this:
Screenshot 2025-06-30 at 07 04 03

apps/web/utils/reply-tracker/enable.ts
edulelis: Yeah. We do not. Removing the explicit `delayInMinutes` from this section.

elie222:
Conflict btw

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -857,7 +859,7 @@ export function RuleForm({
               variant="secondary"
               size="sm"
               onClick={() => {
-                append({ type: ActionType.LABEL });
+                append({ type: ActionType.LABEL, delayInMinutes: null });
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: remove delayinminutes: null. just keep it emoty like all the other fields we could have here but are not set

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -1132,6 +1138,14 @@ function ActionCard({
                       emailAccountId={emailAccountId}
                     />
                   </div>
+                ) : field.name === "label" && isAiGenerated ? (
+                  <div className="mt-2">
+                    <input
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: we can't use the <Input /> component here? is there a reason we're creating a custom one here?

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -1226,6 +1240,68 @@ function ActionCard({
           {/* Show the variable pro tip when action has visible fields that can use variables */}
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: we can drop this comment. feel like ai likes to comment the code, but we don't need it a lot of the time

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -1226,6 +1240,68 @@ function ActionCard({
...
+            </div>
+          )}
+
+          {/* Hidden input for form registration */}
+          {canActionBeDelayed(action.type) && (
+            <input
+              type="hidden"
+              {...register(`actions.${index}.delayInMinutes`)}
+            />
+          )}
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: if it's hidden what's the point of it?

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -1226,6 +1240,68 @@ function ActionCard({
           {/* Show the variable pro tip when action has visible fields that can use variables */}
           {shouldShowProTip && <VariableProTip />}
 
+          {/* Show delay toggle and input for all actions that support delays */}
+          {canActionBeDelayed(action.type) && (
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: think we can call this function higher up. ie `const canBeDelated = ...`

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -1226,6 +1240,68 @@ function ActionCard({
           {/* Show the variable pro tip when action has visible fields that can use variables */}
           {shouldShowProTip && <VariableProTip />}
 
+          {/* Show delay toggle and input for all actions that support delays */}
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: can drop comment

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -1226,6 +1240,68 @@ function ActionCard({
...
+          {canActionBeDelayed(action.type) && (
+            <div className="mt-4">
+              <div className="flex items-start justify-between">
+                <div className="flex-1">
+                  {/* Show delay input controls when delay has been enabled */}
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: drop comment

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -1226,6 +1240,68 @@ function ActionCard({
...
+                      index={index}
+                      delayInMinutes={watch(`actions.${index}.delayInMinutes`)}
+                      setValue={setValue}
+                    />
+                  )}
+                </div>
+
+                <div className="flex items-center space-x-2">
+                  <TooltipExplanation
+                    text="Schedule this action to execute after a delay from when the email was received. Useful for follow-ups, reminders, or giving senders time to respond."
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: Core use case would delayed archive. so think we should adjust this copy a bit.

Comment (elie222) [reply apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: "For example, archive an email a week after receiving it" instead of `Useful for...`

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -1226,6 +1240,68 @@ function ActionCard({
...
+                    enabled={watch(`actions.${index}.delayInMinutes`) !== null}
+                    onChange={(enabled: boolean) => {
+                      if (enabled) {
+                        setValue(`actions.${index}.delayInMinutes`, 60, {
+                          shouldValidate: true,
+                        });
+                      } else {
+                        setValue(`actions.${index}.delayInMinutes`, null, {
+                          shouldValidate: true,
+                        });
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: Simpler:
```ts
setValue(`actions.${index}.delayInMinutes`, enabled ? 60 : null, {
                          shouldValidate: true,
                        });
                        ```

File: apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx [hunk 17]
```diff
@@ -1226,6 +1240,68 @@ function ActionCard({
...
+              </div>
+
+              {errors?.actions?.[index]?.delayInMinutes && (
+                <div className="mt-2">
+                  <ErrorMessage
+                    message={
+                      errors.actions?.[
+                        index
+                      ]?.delayInMinutes?.message?.toString() ||
+                      "Please enter a valid delay between 1 minute and 90 days"
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: zod can handle this error message for us. It will automatically, but if you want to customise the text you can do so too. No need to do it here.

More: https://zod.dev/error-formatting

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -1393,3 +1470,111 @@ function VariableProTip() {
...
+        // Hours (60 minutes = 1 hour)
+        return { value: (minutes / 60).toString(), unit: "hours" };
+      } else {
+        // Minutes
+        return { value: minutes.toString(), unit: "minutes" };
+      }
+    };
+
+    // Convert display value and unit to minutes
+    const convertToMinutes = (value: string, unit: string) => {
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: i think you can pull these helper functions out and put them at the bottom of the file. will make the useMemo much more concise and readble.

can likely drop most of the comments here too.

apps/web/app/(app)/[emailAccountId]/assistant/onboarding/CategoriesSetup.tsx
```diff
@@ -207,6 +207,9 @@ function CategoryCard({
                     <SelectItem value="label_archive">
                       Label + Skip Inbox
                     </SelectItem>
+                    <SelectItem value="label_archive_1_week">
+                      Label + Archive after a week
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/onboarding/CategoriesSetup.tsx]: thinking if we need to make the upper/lowercase more consistent here. ask cursor what it thinks works best

apps/web/app/(app)/admin/page.tsx
```diff
@@ -24,6 +26,12 @@ export default async function AdminPage() {
       <TopSection title="Admin" />
 
       <div className="m-8 space-y-8">
+        <div className="mb-8 flex gap-4">
+          <Link href="/admin/scheduled-actions">
+            <Button variant="outline">View Scheduled Actions</Button>
```
elie222: we'll remove admin functionality from the repo

apps/web/app/api/admin/scheduled-actions/[id]/cancel/route.ts
Comment (elie222) [reply apps/web/app/api/admin/scheduled-actions/[id]/cancel/route.ts]: we're deleting this anyway, but:
use server actions for mutations.

also if you're doing POST and need automatic auth/error handling, then use withAuth, or withError

apps/web/app/api/scheduled-actions/execute/route.ts
```diff
@@ -0,0 +1,121 @@
...
+const logger = createScopedLogger("qstash-scheduled-actions-executor");
+
+export const maxDuration = 300; // 5 minutes
+
+const ScheduledActionPayloadSchema = z.object({
+  scheduledActionId: z.string().min(1, "Scheduled action ID is required"),
+  emailAccountId: z.string().min(1, "Email account ID is required"),
+  messageId: z.string().min(1, "Message ID is required"),
+  threadId: z.string().min(1, "Thread ID is required"),
+  actionType: z.string().min(1, "Action type is required"),
```
elie222: dont think we need any of these messages. it's not user facing anyway).

`z.string()` is fine for all of these.
for `actionType: z.enum()`

But also are you sure we need all these fields, and can't just pass `scheduledActionId` and then fetch the rest of the info from the db?
but if there's a good reason (like better perf) to keep like this then it's fine

apps/web/app/api/scheduled-actions/execute/route.ts
```diff
@@ -0,0 +1,121 @@
...
+export const POST = verifySignatureAppRouter(
+  withError(async (request: NextRequest) => {
+    try {
+      logger.info("QStash request received", {
+        url: request.url,
+        method: request.method,
+        headers: Object.fromEntries(request.headers.entries()),
+      });
+
+      // Parse and validate the request payload
```
elie222: can drop comments

apps/web/app/api/scheduled-actions/execute/route.ts
```diff
@@ -0,0 +1,121 @@
...
+import { markQStashActionAsExecuting } from "@/utils/scheduled-actions/scheduler";
+import { executeScheduledAction } from "@/utils/scheduled-actions/executor";
+import prisma from "@/utils/prisma";
+import { ScheduledActionStatus } from "@prisma/client";
+
+const logger = createScopedLogger("qstash-scheduled-actions-executor");
+
+export const maxDuration = 300; // 5 minutes
+
+const ScheduledActionPayloadSchema = z.object({
```
elie222: => `scheduledActionPayloadSchema`
to be consistent with the project

or even more:
=> `scheduledActionBody` is the format we use in most other places. but up to you.

apps/web/app/api/scheduled-actions/execute/route.ts
```diff
@@ -0,0 +1,121 @@
...
+import type { NextRequest } from "next/server";
+import { z } from "zod";
+import { withError } from "@/utils/middleware";
+import { createScopedLogger } from "@/utils/logger";
+import { markQStashActionAsExecuting } from "@/utils/scheduled-actions/scheduler";
+import { executeScheduledAction } from "@/utils/scheduled-actions/executor";
+import prisma from "@/utils/prisma";
+import { ScheduledActionStatus } from "@prisma/client";
+
+const logger = createScopedLogger("qstash-scheduled-actions-executor");
```
elie222: we don't need qstash mentioned here. this is fine: `scheduled-actions-executor`

apps/web/components/ui/tooltip.tsx
```diff
@@ -6,26 +6,32 @@ import * as TooltipPrimitive from "@radix-ui/react-tooltip";
...
 const TooltipProvider = TooltipPrimitive.Provider;
-
 const Tooltip = TooltipPrimitive.Root;
-
 const TooltipTrigger = TooltipPrimitive.Trigger;
-
+const TooltipPortal = TooltipPrimitive.Portal;
 const TooltipContent = React.forwardRef<
-  React.ElementRef<typeof TooltipPrimitive.Content>,
+  React.ComponentRef<typeof TooltipPrimitive.Content>,
```
elie222: copy the latest shadcn code for this instead. they now use portal. click manual to copy in the code. or use shadcn cli to readd this component:
https://ui.shadcn.com/docs/components/tooltip

apps/web/prisma/migrations/20250626043046_add_scheduled_actions/migration.sql
```diff
@@ -0,0 +1,67 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `digestScheduleId` on the `EmailAccount` table. All the data in the column will be lost.
```
elie222: i would add 1 migration just for that.
then another migration for the actual changes here.

apps/web/prisma/schema.prisma
```diff
@@ -119,36 +119,37 @@ model EmailAccount {
   autoCategorizeSenders Boolean           @default(false)
 
   // Digest email settings
```
elie222: can drop comment

apps/web/prisma/schema.prisma
```diff
@@ -343,6 +344,8 @@ model Action {
   cc      String?
   bcc     String?
   url     String?
+
+  delayInMinutes Int? // Optional delay in minutes (1 min to 30 days)
```
elie222: dont need a new paragraph here. can delete comment

apps/web/prisma/schema.prisma
```diff
@@ -418,10 +422,46 @@ model ExecutedAction {
...
+
+  label   String?
+  subject String?
+  content String?
+  to      String?
+  cc      String?
+  bcc     String?
+  url     String?
+
+  qstashMessageId String? // QStash message ID for efficient cancellation
```
elie222: can we call this something other than messageId? just because we use messageId a lot in the app with a different meaning. i wonder if we even want to mention qstash here.

also no need to repeat the name of the variable in the comment if it's the same thing

apps/web/prisma/schema.prisma
```diff
@@ -418,10 +422,46 @@ model ExecutedAction {
...
+
+  executedAt       DateTime?
+  executedActionId String?   @unique
+
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
```
elie222: these are always the first fields in all other models. let's be consistent and move it up

apps/web/prisma/schema.prisma
```diff
@@ -418,10 +422,46 @@ model ExecutedAction {
...
+
+  executedRule   ExecutedRule    @relation(fields: [executedRuleId], references: [id], onDelete: Cascade)
+  executedAction ExecutedAction? @relation(fields: [executedActionId], references: [id])
+  emailAccount   EmailAccount    @relation(fields: [emailAccountId], references: [id], onDelete: Cascade)
+
+  @@index([status, scheduledFor])
+  @@index([emailAccountId, messageId])
+  @@index([qstashMessageId])
```
elie222: just to check, we need all these?

apps/web/prisma/schema.prisma
```diff
@@ -851,4 +891,12 @@ enum DigestStatus {
...
\ No newline at end of file
+}
+
+enum ScheduledActionStatus {
+  PENDING
+  EXECUTING
+  COMPLETED
+  FAILED
+  CANCELLED
+}
```
elie222, line 898: do we use all states?

apps/web/utils/actions/rule.validation.ts
```diff
@@ -163,7 +169,12 @@ export type UpdateRuleSettingsBody = z.infer<typeof updateRuleSettingsBody>;
 export const enableDraftRepliesBody = z.object({ enable: z.boolean() });
 export type EnableDraftRepliesBody = z.infer<typeof enableDraftRepliesBody>;
 
-const categoryAction = z.enum(["label", "label_archive", "none"]);
+const categoryAction = z.enum([
+  "label",
+  "label_archive",
+  "label_archive_1_week",
```
elie222: wonder if we call this `label_delay_archive`/ `label_delayed_archive` / `label_archive_delayed` or similar. rather than mentioning 1 week which could change if we adjust the code

apps/web/utils/ai/assistant/chat.ts
```diff
@@ -91,6 +92,14 @@ const updateRuleActionsSchema = z.object({
...
         subject: z.string().nullish(),
       }),
+      delayInMinutes: z
+        .number()
+        .min(1, "Minimum supported delay is 1 minute")
+        .max(NINETY_DAYS_MINUTES, "Maximum supported delay is 90 days")
+        .nullish()
+        .describe(
+          "Optional delay in minutes (1 min to 90 days) before executing this action",
+        ),
```
elie222: you seem to have a similar zodtype  twice in the code. are you sure we need both copy pasted?
also note the messages only important if we the user may read it

apps/web/utils/ai/choose-rule/run-rules.ts
```diff
@@ -95,6 +99,24 @@ async function executeMatchedRule(
...
+    });
+  }
+
+  // Separate immediate and delayed actions
+  const immediateActions = actionItems.filter(
+    (item) => !item.delayInMinutes || item.delayInMinutes <= 0,
+  );
+  const delayedActions = actionItems.filter(
+    (item) => item.delayInMinutes != null && item.delayInMinutes > 0,
+  );
```
elie222: lodash `groupBy` may be cleaner here?

apps/web/utils/ai/rule/create-rule-schema.ts
```diff
@@ -76,6 +77,14 @@ const actionSchema = z.object({
...
       "The fields to use for the action. Static text can be combined with dynamic values using double braces {{}}. For example: 'Hi {{sender's name}}' or 'Re: {{subject}}' or '{{when I'm available for a meeting}}'. Dynamic values will be replaced with actual email data when the rule is executed. Dynamic values are generated in real time by the AI. Only use dynamic values where absolutely necessary. Otherwise, use plain static text. A field can be also be fully static or fully dynamic.",
     ),
+  delayInMinutes: z
+    .number()
+    .min(1, "Minimum supported delay is 1 minute")
+    .max(NINETY_DAYS_MINUTES, "Maximum supported delay is 90 days")
+    .nullish()
+    .describe(
+      "Optional delay in minutes (1 min to 90 days) before executing this action",
+    ),
```
elie222: 3rd time we have this code. and we don't need the validation messages here anyway

apps/web/utils/delayed-actions.ts
```diff
@@ -0,0 +1,21 @@
...
+  ActionType.FORWARD,
+  ActionType.DRAFT_EMAIL,
+  ActionType.CALL_WEBHOOK,
+  ActionType.MARK_READ,
+];
+
+/**
+ * Check if an action type supports delayed execution
+ * This is a client-safe utility function
+ */
```
elie222: we don't need these comments. quite clear what the func does based on its name

apps/web/utils/ai/types.ts
```diff
@@ -22,4 +22,5 @@ export type ActionItem = {
   cc?: ExecutedAction["cc"];
   bcc?: ExecutedAction["bcc"];
   url?: ExecutedAction["url"];
+  delayInMinutes?: number | null; // Optional delay in minutes for scheduled actions
```
elie222: we dont need the comments

apps/web/utils/actions/rule.validation.ts
edulelis: I added the 1_week option because the code explicitly delays the archive by one week, as there is no user input. However, I also agree that it makes sense to have `label_archive_delayed` and later allow the user to input the desired delay if needed.

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
Comment (edulelis) [reply apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: This input is actually already like this in the main branch
https://github.com/elie222/inbox-zero/blob/053574e6ec4f8b0770dbeb6c45f12b5a22f9f32d/apps/web/app/(app)/%5BemailAccountId%5D/assistant/RuleForm.tsx#L1182-L1186

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
Comment (edulelis) [reply apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: I believe this input is to cover the use case of:
- Delay was ON, and a time was set
- User sets the delay as OFF.
- Then we must still send the delayInMinutes as NULL or 0 to the backend so it gets updated.

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
Comment (edulelis) [reply apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: Updated

apps/web/app/(app)/[emailAccountId]/assistant/onboarding/CategoriesSetup.tsx
Comment (edulelis) [reply apps/web/app/(app)/[emailAccountId]/assistant/onboarding/CategoriesSetup.tsx]: Sentence case would work best for readability and modern UI patterns:
Option 1: Sentence case (recommended)
"Label"
"Label + skip inbox"
"Label + archive after a week"
"None"
Option 2: Title case (more formal)
"Label"
"Label + Skip Inbox"
"Label + Archive After a Week"
"None"
Sentence case tends to be:
More readable for longer phrases
Less visually "heavy"
More common in modern interfaces
Easier to scan


I refactored to use lowercase

apps/web/app/api/scheduled-actions/execute/route.ts
edulelis: Yes, passing only the `scheduledActionId` is also possible. Going to refactor.

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
Comment (edulelis) [reply apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: Updated and removed this input.

apps/web/prisma/migrations/20250626043046_add_scheduled_actions/migration.sql
edulelis: Needed to execute the migrations once before generating the one for scheduled actions, so it generated in duplicate.

apps/web/prisma/schema.prisma
edulelis: I'll rename this to `queueId`, however in qstash side this is called exactly `messageId`

elie222: hmmm okay. if it's not a queue then probs bad to rename it something it isn't. ask cursor what it thinks for best name, but decide yourself what you think is best

apps/web/prisma/schema.prisma
edulelis: The qstash one is not needed, so I removed it.

apps/web/prisma/schema.prisma
edulelis: Yeah, we do.

PENDING - Initial state, ready for execution
EXECUTING - Lock state, prevents duplicates
COMPLETED - Success state, audit trail
FAILED - Error state, debugging
CANCELLED - Cancelled state, superseded actions (reexecuted an action that had one scheduled action already)

apps/web/utils/ai/assistant/chat.ts
edulelis: Refactored to reuse the zod schema for delayInMinutes.
"Minimum supported delay is 1 minute" can appear only if the user manages to input something below 1 minute (which the UI doesn't really let them)
"Maximum supported delay is 90 days" can appear if the user sets a value than 90 days (unlikely but can happen)

I removed the describe.

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -979,7 +1002,7 @@ function ActionCard({
   errors: any;
   userLabels: NonNullable<LabelsResponse["labels"]>;
   isLoading: boolean;
-  mutate: () => void;
+  mutate: (data?: any) => void;
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: can we avoid any type here

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -91,6 +91,8 @@ import {
 } from "@/components/ui/select";
 import { Form, FormControl, FormField, FormItem } from "@/components/ui/form";
 import { isDefined } from "@/utils/types";
+import { canActionBeDelayed } from "@/utils/delayed-actions";
+import { mutate } from "swr";
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: is this used at all? or just an old import?

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -265,6 +269,24 @@ export function RuleForm({
...
+            delayInMinutes: action.delayInMinutes,
+            content: {
+              ...action.content,
+              setManually: !!action.content?.value,
+            },
+          })),
+        ],
+      });
+    }
+  }, [rule, reset]);
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: this is quite ugly. in general i try to avoid `useEffect reset()`.
what are you trying to acheive here?

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -996,6 +1019,14 @@ function ActionCard({
       ? !!watch(`actions.${index}.content.setManually`)
       : false;
 
+  const actionCanBeDelayed = useMemo(
+    () => canActionBeDelayed(action.type),
+    [action.type],
+  );
+
+  const delayValue = watch(`actions.${index}.delayInMinutes`);
+  const delayEnabled = delayValue !== null && delayValue !== undefined;
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: can we do `const delayEnabled = !!delayValue` instead?
`delayValue = 0` also means it's disabled, right?

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -1095,25 +1126,28 @@ function ActionCard({
...
-                            `actions.${index}.${field.name}`,
-                            enabled
-                              ? { value: "", ai: true }
-                              : { value: "", ai: false },
-                          );
-                        }}
-                      />
-                    </div>
-                  )}
+                  {field.name === "label" &&
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: why the need for the changes in this block of code?

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -1132,6 +1166,15 @@ function ActionCard({
                       emailAccountId={emailAccountId}
                     />
                   </div>
+                ) : field.name === "label" && isAiGenerated ? (
+                  <div className="mt-2">
+                    the field
+                    <input
+                      className="block w-full flex-1 rounded-md border border-border bg-background shadow-sm focus:border-black focus:ring-black sm:text-sm"
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: this won't in dark mode (you can toggle dark mode on in the top right hand corner). you can use things like bg-primary to achieve the same. ask cursor to fix and it should handle it

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -1222,9 +1265,50 @@ function ActionCard({
...
+                  />
+                </div>
+              </div>
+
+              {errors?.actions?.[index]?.delayInMinutes && (
+                <div className="mt-2">
+                  <ErrorMessage
+                    message={errors.actions?.[
+                      index
+                    ]?.delayInMinutes?.message?.toString()}
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: `toString()` needed here?

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -1393,3 +1478,105 @@ function VariableProTip() {
...
+function DelayInputControls({
+  index,
+  delayInMinutes,
+  setValue,
+}: {
+  index: number;
+  delayInMinutes: number | null | undefined;
+  setValue: ReturnType<typeof useForm<CreateRuleBody>>["setValue"];
+}) {
+  const delayConfig = useMemo(() => {
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: bit of a strange pattern to wrap all this in useMemo.

usually each function would be `const func = useCallback(() => {})`
or just: `const func = () => {}`

apps/web/app/api/scheduled-actions/execute/route.ts
```diff
@@ -0,0 +1,109 @@
+import { verifySignatureAppRouter } from "@upstash/qstash/nextjs";
+import type { NextRequest } from "next/server";
+import { z } from "zod";
+import { withError } from "@/utils/middleware";
+import { createScopedLogger } from "@/utils/logger";
+import { markQStashActionAsExecuting } from "@/utils/scheduled-actions/scheduler";
+import { executeScheduledAction } from "@/utils/scheduled-actions/executor";
+import prisma from "@/utils/prisma";
+import { ScheduledActionStatus } from "@prisma/client";
```
elie222, line 9: => `import type { ScheduledActionStatus } from "@prisma/client";`. we should probably add biome to the project to do this automatically. i have it running locally so these errors.

apps/web/hooks/useRule.tsx
```diff
@@ -4,5 +4,9 @@ import type { RuleResponse } from "@/app/api/user/rules/[id]/route";
 export function useRule(ruleId?: string | null) {
   return useSWR<RuleResponse, { error: string }>(
     ruleId ? `/api/user/rules/${ruleId}` : null,
+    {
+      revalidateOnFocus: false,
+      revalidateOnReconnect: false,
+    },
```
elie222: why did we add this? because it resets the page when the user switches tabs?

was that also happening before?

apps/web/prisma/schema.prisma
```diff
@@ -851,4 +887,19 @@ enum DigestStatus {
...
+
+enum SchedulingStatus {
+  PENDING
+  SCHEDULED
+  FAILED
```
elie222, line 903: what's the difference between failed and unavailable?

i prefer to keep it as simple as possible, unless we have a real need for all of them

apps/web/utils/ai/choose-rule/run-rules.ts
```diff
@@ -106,12 +127,24 @@ async function executeMatchedRule(
...
-          actionItems,
+          actionItems: immediateActions, // Only save immediate actions as ExecutedActions
           reason,
         },
       );
 
-  const shouldExecute = executedRule && rule.automate;
+  // Schedule delayed actions if any exist
+  if (executedRule && delayedActions.length > 0 && !isTest) {
+    await scheduleDelayedActions({
```
elie222, line 136: should cancel have gone here? when we reschedule?

apps/web/utils/ai/choose-rule/run-rules.ts
```diff
@@ -95,6 +100,22 @@ async function executeMatchedRule(
     gmail,
   });
 
+  // Cancel any existing scheduled actions for this message
+  if (!isTest) {
```
elie222, line 103: what's the reason we cancel here?

apps/web/utils/action-item.ts
```diff
@@ -177,6 +185,7 @@ export function sanitizeActionFields(
     cc: null,
     bcc: null,
     url: null,
+    delayInMinutes: action.delayInMinutes ?? null,
```
elie222: is there a reason this works differently to any other field?

also isn't delayinminutes 0 also like null?

so `|| null` would be better here anyway?

apps/web/hooks/useRule.tsx
edulelis: I can test before/after.
In another part of the code, the revalidation is more frequent.
I added this in an attempt to fix the following issue: 
- After editing and saving the rule inputs inside the modal, I would get stale data when reopening the modal. I tested only with delayed action input.

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
Comment (edulelis) [reply apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: old import, removed

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
Comment (edulelis) [reply apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: Indeed, this is ugly.
If you edit a rule, save it (closing the modal), and edit it immediately afterwards, you see stale data instead of the updated data. It happens very rarely, and sometimes it is even hard to reproduce.
This helps to avoid this issue. But we could do without it, assuming this is unlikely to happen.

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
Comment (edulelis) [reply apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: Indeed, `delayValue = 0` also means disabled.
However, the zod validation will never allow it because the minimum allowed is 1 minute.

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
Comment (edulelis) [reply apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: Replaced these with the native input to get it to work.

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
Comment (edulelis) [reply apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: removed

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
Comment (edulelis) [reply apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx]: removed the memo wrapper and made simplifications to this function

apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx
```diff
@@ -1333,6 +1398,7 @@ export function ThreadsExplanation({ size }: { size: "sm" | "md" }) {
   return (
```
Comment (edulelis) [inline apps/web/app/(app)/[emailAccountId]/assistant/RuleForm.tsx:1397]: Restored this section to reduce the diff

apps/web/prisma/schema.prisma
edulelis: renamed to `scheduledId`

apps/web/prisma/schema.prisma
edulelis: Failed and unavailable have almost the same meaning.
Could probably use only failed for both.
Failed -> request sent and replied with error
Unavailable -> request failed without reply (service is down)

apps/web/app/api/scheduled-actions/execute/route.ts
edulelis: I don't see this error as this import is being used as a value further down in this file.
https://github.com/elie222/inbox-zero/pull/519/files/1dd91c9d4d73e8f913ddf81701b2af243287c3ba#diff-9d9ed6d5fe8d1989269540e1fd2e77ba707deb0d992125c2dd5894d63b64a8d4R62

apps/web/hooks/useRule.tsx
edulelis: Removing this for now in favor of other solutions.

apps/web/utils/ai/choose-rule/run-rules.ts
edulelis: The functionality is equivalent if cancellation happens here.
The idea is to cancel any existing action first to avoid duplicates.
I'll move it to this section so the code is clearer.

apps/web/utils/ai/choose-rule/run-rules.ts
edulelis: The cancellation here was merely a means to separate concerns. 
Cancel everything first -> execute/schedule everything afterwards.
Cancelling immediately before the scheduling also works.

apps/web/utils/action-item.ts
edulelis: This is just because all other fields are set as null, and delayInMinutes could be `undefined`. But yes, `|| null` will also cover the `0`

edulelis:
@coderabbitai full review

PR #523, opened 2025-06-24
Fix lint errors

miha-yy:


apps/web/app/(app)/[emailAccountId]/stats/NewsletterModal.tsx
```diff
@@ -199,7 +200,7 @@ function UnarchivedEmails({
     <LoadingContent loading={isLoading} error={error}>
       {data && (
         <EmailList
-          threads={data.threads}
+          threads={data.threads as Thread[]}
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/stats/NewsletterModal.tsx:203]: do not use` as Thread`. If this was erroring, then the actual error needs to be fixed

apps/web/app/(app)/[emailAccountId]/stats/NewsletterModal.tsx
```diff
@@ -30,6 +30,7 @@ import type { Row } from "@/app/(app)/[emailAccountId]/bulk-unsubscribe/types";
 import { useThreads } from "@/hooks/useThreads";
 import { useAccount } from "@/providers/EmailAccountProvider";
 import { onAutoArchive } from "@/utils/actions/client";
+import { Thread } from "@/components/email-list/types";
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/stats/NewsletterModal.tsx:33]: you can do `import type { Thread } from "@/components/email-list/types";`
but not critical

apps/web/app/(app)/[emailAccountId]/stats/NewsletterModal.tsx
```diff
@@ -231,7 +232,7 @@ function AllEmails({
     <LoadingContent loading={isLoading} error={error}>
       {data && (
         <EmailList
-          threads={data.threads}
+          threads={data.threads as Thread[]} // TODO: FIX FOR OUTLOOK
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/stats/NewsletterModal.tsx:235]: fix

apps/web/components/GroupedTable.tsx
```diff
@@ -555,28 +557,41 @@ function ExpandedRows({
...
+        const firstMessage = thread.messages[0];
+        // Handle both Google (with headers) and Microsoft (direct properties) message structures
+        const subject =
+          "headers" in firstMessage
+            ? firstMessage.headers.subject
+            : firstMessage.subject;
+        const date =
+          "headers" in firstMessage
+            ? firstMessage.headers.date
+            : firstMessage.receivedAt;
```
elie222, line 570: i think we should be returning a unified format on the server rather than lots of one off checks like this.

eg. set message.date, and message.subject, so that you don't need any special logic in this part of the code

apps/web/app/(app)/[emailAccountId]/bulk-unsubscribe/common.tsx
```diff
@@ -50,7 +50,7 @@ import {
   useDeleteAllFromSender,
 } from "@/app/(app)/[emailAccountId]/bulk-unsubscribe/hooks";
 import { LabelsSubMenu } from "@/components/LabelsSubMenu";
-import type { UserLabel } from "@/hooks/useLabels";
+import { EmailLabel } from "@/providers/EmailProvider";
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/bulk-unsubscribe/common.tsx:53]: import type { EmailLabel } from "@/providers/EmailProvider";

I use `biome` Cursor extension that automatically surfaces these. Maybe I add to the project for everyone to use

apps/web/components/email-list/PlanActions.tsx
```diff
@@ -7,14 +7,15 @@ import type { Executing, Thread } from "@/components/email-list/types";
...
+import { ThreadsResponse } from "@/hooks/useThreads";
 
 export function useExecutePlan(refetch: () => void) {
   const [executingPlan, setExecutingPlan] = useState<Executing>({});
   const [rejectingPlan, setRejectingPlan] = useState<Executing>({});
   const { emailAccountId } = useAccount();
 
   const executePlan = useCallback(
-    async (thread: Thread) => {
+    async (thread: ThreadsResponse["threads"][number]) => {
```
elie222, line 18: may want to export the type rather than having to do `ThreadsResponse["threads"][number]` each time

apps/web/hooks/useLabels.ts
```diff
@@ -71,19 +72,25 @@ export function useLabels() {
...
         .map((label) => ({
-          id: label.id,
-          name: label.name,
-          type: "user" as const,
+          id: label.id || "",
+          name: label.name || "",
+          type: label.type || null,
           color: label.color,
         }))
         .sort(sortLabels);
```
elie222, line 96: feels messy that we need to do this here. ideally we have unified apis that are returning consistent data

apps/web/hooks/useThreads.ts
```diff
@@ -3,7 +3,7 @@ import type { ThreadsResponse as GmailThreadsResponse } from "@/app/api/google/t
 import type { ThreadsResponse as MicrosoftThreadsResponse } from "@/app/api/microsoft/threads/controller";
 import { useAccount } from "@/providers/EmailAccountProvider";
 
-type ThreadsResponse = GmailThreadsResponse | MicrosoftThreadsResponse;
+export type ThreadsResponse = GmailThreadsResponse | MicrosoftThreadsResponse;
```
elie222, line 6: i think api itself should be returning a consistent unified format that we define

PR #540, opened 2025-07-04
Bulk unsubscribe

miha-yy:


apps/web/app/api/outlook/webhook/process-history-item.ts
```diff
@@ -284,20 +284,24 @@ export async function processHistoryItem(
     snippet: message.bodyPreview || "",
     historyId: message.id || messageId,
     inline: [],
+    subject: message.subject || "",
+    date: message.receivedDateTime
```
elie222, line 288: is this because of google vs microsoft? can we handle this at provider level and have unified message format instead?

PR #541, opened 2025-07-04
Unify routes and bulk unsubscribe webhook functions

miha-yy:


apps/web/app/api/labels/create/route.ts
```diff
@@ -0,0 +1,44 @@
...
+  const emailAccount = await prisma.emailAccount.findUnique({
+    where: { id: emailAccountId },
+    select: {
+      account: {
+        select: {
+          provider: true,
+        },
+      },
+    },
+  });
```
elie222, line 27: as we do this in a few places, probably good to just make this:

await getProvider() or getAccountProvider()

could even optimise it, and store the result in redis similar to what we do here:
`apps/web/utils/redis/account-validation.ts`

especially as accountid<>provider will never actually change.

we could actually reuse the existing getEmailAccount and have it store an object instead: { email, provider } (instead of just the email string). but to not break production, you need to check if the cached result is an object, and if it isn't, then it's like we have nothing in the cache. and it needs to refetched. alternative is to use a new validation key

apps/web/app/api/labels/create/route.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  if (!emailAccount) {
+    return NextResponse.json(
+      { error: "Email account not found" },
+      { status: 404 },
+    );
+  }
+
+  const provider = emailAccount.account.provider;
+  const emailProvider = await createEmailProvider({ emailAccountId, provider });
```
elie222, line 37: getEmailProvider may be a better name for this function.

another thought: as we're doing this a lot and have the above block of code repeat throughout the app, i wonder if we can put it in the middleware:

POC code (AI generated you don't need to do what it's done 1:1, but should show the idea):

```

export interface RequestWithEmailProvider extends RequestWithEmailAccount {
  emailProvider: EmailProvider;
}

async function emailProviderMiddleware(
  req: NextRequest,
): Promise<RequestWithEmailProvider | Response> {
  // First run email account middleware
  const emailAccountReq = await emailAccountMiddleware(req);
  if (emailAccountReq instanceof Response) return emailAccountReq;

  const { userId, emailAccountId } = emailAccountReq.auth;

  try {
    // Get the email account with its related account to get the provider
    const emailAccount = await prisma.emailAccount.findUnique({
      where: { 
        id: emailAccountId,
        userId // ensure it belongs to the user
      },
      include: {
        account: {
          select: {
            provider: true
          }
        }
      }
    });
    
    if (!emailAccount) {
      return NextResponse.json(
        { error: "Email account not found", isKnownError: true },
        { status: 404 },
      );
    }

    // Create the email provider
    const provider = await createEmailProvider({
      emailAccountId: emailAccount.id,
      provider: emailAccount.account.provider,
    });
    
    // Create a new request with email provider
    const providerReq = emailAccountReq.clone() as RequestWithEmailProvider;
    providerReq.emailProvider = provider;
    
    return providerReq;
  } catch (error) {
    logger.error("Failed to create email provider", {
      error,
      emailAccountId,
      userId,
    });
    return NextResponse.json(
      { error: "Failed to initialize email provider", isKnownError: true },
      { status: 500 },
    );
  }
}

export function withEmailProvider(
  handler: NextHandler<RequestWithEmailProvider>,
): NextHandler {
  return withMiddleware(handler, emailProviderMiddleware);
}
```

and then the code in many files would be:

```
export const GET = withEmailProvider(
  async (req: RequestWithEmailProvider) => {
    const { emailProvider } = req;
    // Now you can use emailProvider to interact with Gmail or Outlook
    const messages = await emailProvider.getMessages();
    // ...
  }
);
```

apps/web/app/api/threads/validation.ts
```diff
@@ -6,6 +6,7 @@ export const threadsQuery = z.object({
   type: z.string().nullish(),
   q: z.string().nullish(),
   nextPageToken: z.string().nullish(),
-  labelId: z.string().nullish(),
+  labelId: z.string().nullish(), // For Google
+  folderId: z.string().nullish(), // For Microsoft
```
elie222, line 10: could we just make this labelId to handle both? and then the provider handles each as needed later on

PR #548, opened 2025-07-08
Email display subject instead of whole email

miha-yy:


elie222:
Nice. If you have a screenshot for this will be helpful!

PR #568, opened 2025-07-14
Add biome via ultracite

elie222:


elie222:
Replaced with: https://github.com/elie222/inbox-zero/pull/569

PR #583, opened 2025-07-20
chore: update gitignore and reduce maxDuration in admin page

finaldesignstudios:
Add CONTEXT/ to gitignore and decrease maxDuration from 800 to 300 seconds in admin page for better resource management and to comply with Vercel hobbyPlan max limit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,9 @@ dist
 **/public/precache**
 
 # Sanity
+# Ignore context directory
+CONTEXT/
+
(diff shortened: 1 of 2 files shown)


elie222:
Hey, I'm aware of this one and wish Vercel would do this better. You can turn Fluid Compute on btw. But if not then just manually change from 800 to 300 on your end. I tried to make it an env var in the past, but Vercel doesn't support that.

PR #589, opened 2025-07-23
fix: Fix outlook thread detection

edulelis:


apps/web/utils/ai/choose-rule/match-rules.ts
```diff
@@ -1,4 +1,3 @@
-import type { gmail_v1 } from "@googleapis/gmail";
```
edulelis, line 1: Removed unused imports

apps/web/utils/ai/choose-rule/match-rules.ts
```diff
@@ -234,7 +231,7 @@ async function findMatchingRuleWithReasons(
   matchReasons?: MatchReason[];
   reason?: string;
 }> {
-  const isThread = isReplyInThread(message.id, message.threadId);
+  const isThread = message.isReplyInThread;
```
edulelis, line 234: Moved this to a param instead so each email provider can handle it independently.
Gmail uses message.id and message.threadId
Outlook uses message.conversationIndex

apps/web/utils/outlook/message.ts
```diff
@@ -1,4 +1,3 @@
-import type { Client } from "@microsoft/microsoft-graph-client";
```
edulelis, line 1: Removed unused import

apps/web/utils/ai/choose-rule/match-rules.ts
elie222: ok. not sure there's a need for each time as it's not used often, but ill merge as is

PR #607, opened 2025-07-30
Create stats API

elie222:


elie222:
Added here: https://github.com/elie222/inbox-zero/pull/1085

PR #608, opened 2025-07-30
fix: Fix old drafts not being deleted

edulelis:


apps/web/utils/ai/actions.ts
```diff
@@ -114,6 +114,7 @@ const draft: ActionFunction<{
       labelIds: [],
     },
     draftArgs,
+    executedRule,
```
edulelis, line 117: This param was missing so it never triggered deleting old drafts.

apps/web/utils/email/provider.ts
```diff
@@ -440,10 +440,10 @@ export class GmailProvider implements EmailProvider {
           logger,
         }),
       ]);
-      return { draftId: result.data.message?.id || "" };
+      return { draftId: result.data.id || "" };
```
edulelis, line 443: The `draftId` is not the id inside the message object.
Deletion would fail afterwards because we did not store the correct `draftId`

apps/web/utils/gmail/draft.ts
```diff
@@ -28,7 +28,7 @@ export async function deleteDraft(gmail: gmail_v1.Gmail, draftId: string) {
       userId: "me",
       id: draftId,
     });
-    if (response.status !== 200) {
+    if (response.status !== 200 && response.status !== 204) {
```
edulelis, line 31: The response here is `204 No Content`, so we shouldn't log it as an error.

PR #609, opened 2025-07-30
fix: Move all extra Microsoft Auth data to the database

edulelis:
## Summary by CodeRabbit

* **New Features**
  * Enhanced Microsoft OAuth sign-in experience with improved session data, including additional Microsoft-specific information.
  * Optimized token handling for Microsoft users to reduce cookie size while maintaining secure session data.

### DIFF
--- a/apps/web/utils/auth.ts
+++ b/apps/web/utils/auth.ts
@@ -223,6 +223,8 @@ export const getAuthOptions: () => NextAuthConfig = () => ({
       return true;
     },
     jwt: async ({ token, user, account }): Promise<JWT> => {
+      // Signing in
+      // on first sign in `account` and `user` are defined, thereafter only `token` is defined
       if (account && user) {
         if (account.provider === "microsoft-entra-id") {
           if (account.refresh_token) {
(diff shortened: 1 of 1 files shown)


apps/web/utils/auth.ts
```diff
@@ -210,70 +210,102 @@ export const getAuthOptions: () => NextAuthConfig = () => ({
   // based on: https://authjs.dev/guides/basics/refresh-token-rotation
   // and: https://github.com/nextauthjs/next-auth-refresh-token-example/blob/main/pages/api/auth/%5B...nextauth%5D.js
   callbacks: {
+    signIn: async ({ user, account }) => {
```
elie222: Is this needed now?

apps/web/utils/auth.ts
```diff
@@ -654,6 +727,12 @@ declare module "next-auth" {
     user: {} & DefaultSession["user"] & { id: string };
     accessToken?: string;
     error?: string | "RefreshAccessTokenError";
+    microsoftData?: {
```
elie222: Do we need this?

apps/web/utils/auth.ts
edulelis: This can be removed.

apps/web/utils/auth.ts
```diff
@@ -289,15 +286,36 @@ export const getAuthOptions: () => NextAuthConfig = () => ({
...
+
+        if (user) {
+          session.user = {
+            id: user.id,
+            email: user.email,
+            name: user.name,
+            image: user.image,
+          };
+        }
+      }
```
elie222, line 314: why do we need image in the session though?  we'll just fetch the image when we actually need it. are you sure we need this db call at all?

PR #638, opened 2025-08-05
Update migrations

edulelis:


apps/web/prisma/migrations/20250804163003_better_auth/migration.sql
johnlowe399-blip, line 1: Ok can you send me phrase code

PR #672, opened 2025-08-12
fix: Update schemas do disallow move to folder from rule prompt creation

edulelis:


apps/web/utils/actions/assess.ts
```diff
@@ -46,6 +46,7 @@ export const analyzeWritingStyleAction = actionClient
         email: true,
         about: true,
         user: { select: { aiProvider: true, aiModel: true, aiApiKey: true } },
+        account: { select: { provider: true } },
```
elie222, line 49: Provider is passed in the context of the action already

apps/web/utils/actions/rule.ts
```diff
@@ -426,6 +426,7 @@ export const deleteRuleAction = actionClient
                 aiApiKey: true,
               },
             },
+            account: { select: { provider: true } },
```
elie222, line 429: Provider is passed in the context of the action already

apps/web/utils/ai/rule/create-rule-schema.ts
```diff
@@ -33,70 +34,81 @@ const conditionSchema = z
...
+        webhookUrl: z
+          .string()
+          .nullish()
+          .transform((v) => v ?? null)
+          .describe("The webhook URL to call"),
+        folderName: z
+          .string()
+          .nullish()
+          .transform((v) => v ?? null)
+          .describe("The folder to move the email to"),
```
elie222, line 89: it shouldn't be in the schema at all when provider == google

elie222:
If you can fix up the notes after. Merging now as nothing critical.

PR #678, opened 2025-08-13
Refactor Outlook folder functionality

edulelis:
Screenshot 2025-08-13 at 12 05 44

Screenshot 2025-08-13 at 12 04 46

## Summary by CodeRabbit

* **New Features**
  * Hierarchical, searchable folder picker and combobox for selecting Outlook folders (with loading/error states) and per-action folder selection UI.

* **Improvements**
  * Actions use stable folder IDs for move/archive operations; folder fields disable variable insertion and pro-tips.
  * Provider-aware rule handling: folder fields are included only for Microsoft accounts; provider threaded through rule tools.

* **Bug Fixes / Validation**
  * MOVE_FOLDER validation accepts a selected folder ID or name and shows clearer prompts.

* **Chores**
  * Backend and schema updated to store folder IDs; folder list API and client hook added.


apps/web/app/api/user/folders/route.ts
```diff
@@ -0,0 +1,44 @@
...
+      account: {
+        select: {
+          provider: true,
+          access_token: true,
+          refresh_token: true,
+          expires_at: true,
+        },
+      },
+    },
+  });
```
elie222: we have a getter function in the project that gets this stuff btw:
getEmailAccountWithAiAndTokens

or could add another getter in `apps/web/utils/user/get.ts` file that's only for tokens perhaps

apps/web/hooks/useFolders.ts
```diff
@@ -0,0 +1,12 @@
+import type { OutlookFolder } from "@/utils/outlook/folders";
+import useSWR from "swr";
+
+export function useFolders({ emailAccountId }: { emailAccountId: string }) {
```
elie222: You don't need to pass emailAccountId.

See useRules. It works because of this in SWRProvider

```ts
// https://swr.vercel.app/docs/error-handling#status-code-and-error-object
const fetcher = async (
  url: string,
  init?: RequestInit | undefined,
  emailAccountId?: string | null,
) => {
  const headers = new Headers(init?.headers);

  if (emailAccountId) {
    headers.set(EMAIL_ACCOUNT_HEADER, emailAccountId);
  }
```

elie222: Actually I don't think the code is using your param at all. The backend doesn't process it.

I would switch this file to be like the other swr hook files. Super simple:

```ts
import useSWR from "swr";
import type { RulesResponse } from "@/app/api/user/rules/route";

export function useRules() {
  return useSWR<RulesResponse, { error: string }>("/api/user/rules");
}
```

apps/web/app/api/user/folders/route.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  if (emailAccount?.account?.provider === "microsoft") {
+    const client = await getOutlookClientWithRefresh({
+      accessToken: emailAccount.account.access_token,
+      refreshToken: emailAccount.account.refresh_token,
+      expiresAt: emailAccount.account.expires_at
+        ? Math.floor(emailAccount.account.expires_at.getTime() / 1000)
+        : null,
```
elie222: we shouldn't have to do this each time. getOutlookClientWithRefresh should be handling this. 

I'm seeing other parts of the codebase that look this which isn't consistent and means there's a bug somewhere:

```ts
    const outlookClient = await getOutlookClientWithRefresh({
      accessToken: emailAccount.account?.access_token,
      refreshToken: emailAccount.account?.refresh_token,
      expiresAt: emailAccount.account?.expires_at?.getTime() || null,
      emailAccountId: emailAccount.id,
    });
```

elie222: This is the code I'd expect and then `getOutlookClientWithRefresh` handles the expires at correctly:

```ts
const outlookClient = await getOutlookClientWithRefresh({
      accessToken: emailAccount.account?.access_token,
      refreshToken: emailAccount.account?.refresh_token,
      expiresAt: emailAccount.account?.expires_at,
      emailAccountId: emailAccount.id,
    });
```

apps/web/app/api/user/folders/route.ts
edulelis: Going to refactor.
The referenced branch on which this was built did not have `provider` in `getEmailAccountWithAiAndTokens`.

apps/web/hooks/useFolders.ts
edulelis: Ok. Let me refactor.
CodeRabbit added some feedback saying this wouldn't work.

apps/web/app/api/user/folders/route.ts
edulelis: Going to replace it with `getOutlookClientForEmail` as it already calls `getOutlookClientWithRefresh`

apps/web/utils/ai/choose-rule/ai-choose-rule.ts
```diff
@@ -85,7 +85,9 @@ ${emailSection}
     system,
     prompt,
     schema: z.object({
-      reason: z.string().describe("The reason you chose the rule. Keep it concise"),
+      reason: z
```
edulelis, line 88: commit hook formatting

apps/web/utils/actions/rule.ts
```diff
@@ -392,62 +396,64 @@ export const enableDraftRepliesAction = actionClient
...
-    if (rule.emailAccountId !== emailAccountId)
-      throw new SafeError("You don't have permission to delete this rule");
-
-    try {
-      await deleteRule({
-        ruleId: id,
-        emailAccountId,
-        groupId: rule.groupId,
+  .action(
+    async ({ ctx: { emailAccountId, provider }, parsedInput: { id } }) => {
```
edulelis, line 400: commit hook formatting

apps/web/utils/reply-tracker/enable.ts
```diff
@@ -137,6 +137,15 @@ export async function createToReplyRule(
   emailAccountId: string,
   addDigest: boolean,
 ) {
+  const emailAccount = await prisma.emailAccount.findUnique({
```
elie222, line 140: 1. You can just pass in provider. I'd really think before making db calls unnecessarily like this. We're passing provider around all over the app. There's a very high chance the function calling this already has that info.

2. not actually sure how critical it is to pass provider down. i guess it's just to check we don't create a folder rule for gmail users?

apps/web/utils/rule/rule.ts
```diff
@@ -361,7 +376,9 @@ function mapActionFields(
       subject: a.fields?.subject,
       content: a.fields?.content,
       url: a.fields?.webhookUrl,
-      folderName: a.fields?.folderName,
+      ...(provider === "microsoft" && {
+        folderName: a.fields?.folderName as string | null,
```
elie222, line 380: we avoid `as` here?

or `folderName || null` might fix it?

apps/web/utils/outlook/label.ts
```diff
@@ -293,18 +319,21 @@ export async function archiveThread({
...
       throw error;
     }
 
     if (publishResult.status === "rejected") {
-      logger.error(`Failed to publish action to move thread to ${folderName}`, {
-        threadId,
-        error: publishResult.reason,
-      });
+      logger.error(
+        `Failed to publish action to move thread to folder ${folderId}`,
```
elie222, line 331: i've made this note before. if you're adding logs, then don't use variables in it. harder to analyse logs this way

apps/web/utils/outlook/label.ts
```diff
@@ -293,18 +319,21 @@ export async function archiveThread({
         logger.warn("Thread not found", { threadId, userEmail: ownerEmail });
         return { status: 404, message: "Thread not found" };
       }
-      logger.error(`Failed to move thread to ${folderName}`, {
+      logger.error(`Failed to move thread to folder ${folderId}`, {
```
elie222, line 322: log with var

apps/web/utils/outlook/folders.ts
```diff
@@ -0,0 +1,80 @@
...
+    }));
+  };
+
+  const processedFolders = processFolders(folders);
+  for (let currentLevel = 0; currentLevel < expandLevels; currentLevel++) {
+    const fetchNested = async (folderList: OutlookFolder[]) => {
+      for (const folder of folderList) {
+        if (!folder.childFolders || folder.childFolders.length === 0) {
+          try {
+            const childFolders = await getFolders(
```
elie222, line 54: why pass path in? wouldn't it be clearer to do `getChildFolders`?

apps/web/utils/outlook/folders.ts
```diff
@@ -0,0 +1,80 @@
...
+      .select(fields)
+      .expand(
+        `childFolders($select=${fields};$expand=childFolders($select=${fields}))`,
+      )
+      .get();
+
+    return response.value;
+  };
+
+  const folders = await getFolders("/me/mailFolders");
```
elie222, line 34: no need to pass in path. see comment below

apps/web/utils/ai/assistant/chat.ts
```diff
@@ -434,7 +442,7 @@ const updateRuleActionsTool = ({
           bcc: action.bcc,
           subject: action.subject,
           webhookUrl: action.url,
-          folderName: action.folderName,
+          folderName: provider === "microsoft" ? action.folderName : undefined,
```
elie222, line 445: might make sense to have `isMicrosoft` helper instead of magic string each time.

PR #689, opened 2025-08-15
WIP: Learned patterns from labels

edulelis:


apps/web/app/api/google/webhook/process-history.ts
```diff
@@ -232,26 +238,40 @@ async function processHistory(options: ProcessHistoryOptions) {
...
       ...(h.messagesAdded || []),
       ...(h.labelsAdded || []),
+      ...(h.labelsRemoved || []),
     ];
 
     if (!historyMessages.length) continue;
 
-    const inboxMessages = historyMessages.filter(isInboxOrSentMessage);
-    const uniqueMessages = uniqBy(inboxMessages, (m) => m.message?.id);
+    const allEvents = [
```
edulelis, line 243: Adding a message type field here, wrapping the event item, so it is possible to discover the actual event tied to the message in the underlying functions.
Mainly to differentiate between LABEL_ADDED vs LABEL_REMOVED events inside `processHistoryItem`

apps/web/app/api/google/webhook/process-history-item.ts
```diff
@@ -25,11 +25,19 @@ import type { EmailAccountWithAI } from "@/utils/llms/types";
...
+import { inboxZeroLabels } from "@/utils/label";
+import { HistoryEventType } from "@/app/api/google/webhook/process-history";
 
 export async function processHistoryItem(
-  {
-    message,
-  }: gmail_v1.Schema$HistoryMessageAdded | gmail_v1.Schema$HistoryLabelAdded,
+  wrappedItem: {
+    type: HistoryEventType;
+    item:
```
edulelis, line 34: For label-added and label-removed events, we need the labelIds field, which is outside the message object, and in the message-added event, it doesn't exist, so destructuring as `{ message, labelIds }` is not possible.
The extra wrapping is to bring to this function the event that originated the message, so they can be handled separately.

apps/web/app/api/google/webhook/process-history-item.ts
```diff
@@ -52,6 +62,24 @@ export async function processHistoryItem(
...
+  });
+
+  if (type === HistoryEventType.LABEL_REMOVED) {
+    logger.info("Processing label removed event for learning", loggerOptions);
+    return handleLabelRemovedEvent(item, {
+      gmail,
+      emailAccount,
+      provider,
+    });
+  } else if (type === HistoryEventType.LABEL_ADDED) {
```
edulelis, line 75: The label-removed event could be cached for 30 secs ~ 5 minutes, and later, a retrieval of the cache could occur here. This would make it possible to tie both events together.

apps/web/app/api/google/webhook/process-history-item.ts
```diff
@@ -97,11 +120,6 @@ export async function processHistoryItem(
       emailToCheck: parsedMessage.headers.to,
     });
 
-    const provider = await createEmailProvider({
```
edulelis, line 100: provider was being created twice

apps/web/app/api/google/webhook/process-history-item.ts
```diff
@@ -292,3 +310,187 @@ export function shouldRunColdEmailBlocker(
...
+  // Don't learn from To Reply rules
+  // TODO: Add some learning from To Reply rules
+  if (executedRule.rule.systemType === "TO_REPLY") {
+    logger.info("Skipping learning for To Reply rule", loggerOptions);
+    return;
+  }
+
+  logger.info("Saving learned exclusion pattern", loggerOptions);
+  try {
+    await saveLearnedPatterns({
```
edulelis: Landing here means the user removed a label that the AI added.

apps/web/app/api/google/webhook/process-history-item.ts
```diff
@@ -292,3 +310,187 @@ export function shouldRunColdEmailBlocker(
     hasAiAccess
   );
 }
+
+async function handleLabelRemovedEvent(
```
elie222: lets move label removed handling into its own file and import it

apps/web/app/api/google/webhook/process-history-item.ts
elie222: ok. id probably call it historyitem or item instead of wrappeditem

apps/web/app/api/google/webhook/process-history-item.ts
elie222: ya. fine to leave for this first pr. adds more complexity

apps/web/app/api/google/webhook/process-history-item.ts
```diff
@@ -52,6 +62,24 @@ export async function processHistoryItem(
...
+    logger.info("Processing label removed event for learning", loggerOptions);
+    return handleLabelRemovedEvent(item, {
+      gmail,
+      emailAccount,
+      provider,
+    });
+  } else if (type === HistoryEventType.LABEL_ADDED) {
+    logger.info("Processing label added event for learning", loggerOptions);
+    // TODO: Implement handleLabelAddedEvent for learning from user-applied labels
+    return;
```
elie222, line 77: so this changes a lot in this function. it's probably fine as im not sure we were doing anything with it but we should double check that it doesnt break anything.

also we can remove todo and log for the meantime. lets just focus this pr on learning from label removed events

apps/web/app/api/google/webhook/process-history-item.ts
```diff
@@ -292,3 +310,187 @@ export function shouldRunColdEmailBlocker(
...
+    provider: EmailProvider;
+  },
+) {
+  const messageId = message.message?.id;
+  const threadId = message.message?.threadId;
+  const emailAccountId = emailAccount.id;
+  const userEmail = emailAccount.email;
+
+  if (!messageId || !threadId) {
+    logger.info("Skipping label removal - missing messageId or threadId", {
```
elie222: make this a `warn`

apps/web/app/api/google/webhook/process-history-item.ts
```diff
@@ -292,3 +310,187 @@ export function shouldRunColdEmailBlocker(
...
+      threadId,
+    });
+    return;
+  }
+
+  const loggerOptions = {
+    email: userEmail,
+    messageId,
+    threadId,
+  };
```
elie222: this can be moved up as you use the same log option for the previous log.info

apps/web/app/api/google/webhook/process-history-item.ts
```diff
@@ -292,3 +310,187 @@ export function shouldRunColdEmailBlocker(
...
+    const parsedMessage = await provider.getMessage(messageId);
+    const sender = extractEmailAddress(parsedMessage.headers.from);
+
+    // For label removal, we need to get the labelIds from the HistoryLabelRemoved object
+    const removedLabelIds = message.labelIds || [];
+    const removedLabels = await Promise.all(
+      removedLabelIds.map((labelId: string) =>
+        getLabelById({ gmail, id: labelId }),
+      ),
+    );
```
elie222: easier is to just fetch all labels for the user. and then use javascript `.find` to grab what you need

this makes it one call to gmail api instead of potentially multiple (unless we know for sure it's always 1 label here? but then promise.all wouldn't be needed either)

apps/web/app/api/google/webhook/process-history-item.ts
```diff
@@ -292,3 +310,187 @@ export function shouldRunColdEmailBlocker(
...
+  };
+
+  logger.info("Processing label removal for learning", loggerOptions);
+
+  try {
+    const parsedMessage = await provider.getMessage(messageId);
+    const sender = extractEmailAddress(parsedMessage.headers.from);
+
+    // For label removal, we need to get the labelIds from the HistoryLabelRemoved object
+    const removedLabelIds = message.labelIds || [];
```
elie222: note, we don't want to learn from INBOX label removed. that's just archiving emails

same for READ, and may also apply to things like SENT and DRAFT

apps/web/app/api/google/webhook/process-history-item.ts
```diff
@@ -292,3 +310,187 @@ export function shouldRunColdEmailBlocker(
...
+    where: {
+      unique_emailAccount_thread_message: {
+        emailAccountId,
+        threadId,
+        messageId,
+      },
+    },
+    include: {
+      rule: {
+        include: {
```
elie222: minor optimisation is to use `select`'s instead

apps/web/app/api/google/webhook/process-history-item.ts
edulelis: Filtered known system labels

apps/web/app/api/google/webhook/process-label-removed-event.ts
```diff
@@ -0,0 +1,142 @@
...
+    sender,
+  };
+
+  // Can't learn patterns without knowing who to exclude
+  if (!sender) {
+    logger.info("No sender found, skipping learning", loggerOptions);
+    return;
+  }
+
+  if (labelName === inboxZeroLabels.cold_email.name) {
```
edulelis, line 118: Only handling Cold email tag removal for now.

elie222:
@claude check this

elie222:
@claude review

PR #690, opened 2025-08-18
fix: Run filter reply when the reply rule is group matched

edulelis:
## Summary by CodeRabbit

* **New Features**
  * Reply-based rules are now routed through a potential matches step, providing clearer suggestions before a rule is applied.
* **Bug Fixes**
  * Improved accuracy for reply-triggered automations by factoring in sender reply history and thresholds, reducing unintended matches and actions.
* **Refactor**
  * Streamlined rule-matching flow for reply-based scenarios to align with the potential-matches workflow without affecting other rule types.

### DIFF
--- a/apps/web/utils/ai/choose-rule/match-rules.ts
+++ b/apps/web/utils/ai/choose-rule/match-rules.ts
@@ -373,9 +373,7 @@ async function matchesCategoryRule(
 }
 
 export async function filterToReplyPreset(
-  potentialMatches: (RuleWithActionsAndCategories & {
-    instructions: string | null | undefined;
-  })[],
+  potentialMatches: (RuleWithActionsAndCategories & { instructions: string })[],
   message: ParsedMessage,
   client: EmailProvider,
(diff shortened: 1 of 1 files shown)


apps/web/utils/ai/choose-rule/match-rules.ts
```diff
@@ -122,6 +122,15 @@ async function findPotentialMatchingRules({
           group,
         });
 
+        if (rule.systemType === SystemType.TO_REPLY) {
+          const filteredPotentialMatches = await filterToReplyPreset(
```
edulelis, line 126: This should run even when the to reply rule is matched.

elie222:
This doesn't seem to be a fix. As it's only running on the learned patterns section, which we actually want to respect. I'll try check more on my end why it's happening.

edulelis:
> This doesn't seem to be a fix. As it's only running on the learned patterns section, which we actually want to respect. I'll try check more on my end why it's happening.

Ok. I'm going to close this one for now and will open a new one once more info is available.

PR #697, opened 2025-08-19
Learn from label removal patterns for newsletters

edulelis:
## Summary by CodeRabbit

- New Features
  - AI-powered analysis when you remove a label to refine auto-sorting rules.
  - Automatically adds or removes learned patterns based on your feedback.
  - Stores a short “reasoning” note for each learned pattern for clearer context.
  - Broader recognition of system labels (e.g., Sent, Drafts, Spam) to avoid false triggers.
  - Improved handling when removing the “Cold Email” label to keep statuses accurate.

- Tests
  - Added extensive test coverage for AI label-removal analysis scenarios.


apps/web/app/api/google/webhook/process-label-removed-event.ts
```diff
@@ -20,6 +27,22 @@ const SYSTEM_LABELS = [
...
+  return await prisma.rule.findMany({
+    where: {
+      emailAccountId,
+      systemType: {
+        not: null,
+      },
+    },
+    select: {
+      name: true,
+      instructions: true,
```
edulelis: Fetching and passing label instructions to AI prompt

apps/web/utils/ai/label-analysis/analyze-label-removal.ts
```diff
@@ -0,0 +1,129 @@
...
+  });
+
+  const aiResponse = await generateObject({
+    ...modelOptions,
+    system,
+    prompt,
+    schema,
+  });
+
+  logger.trace("Output", aiResponse.object);
```
elie222: i think this is already traced by `createGenerateObject`

apps/web/utils/ai/label-analysis/analyze-label-removal.ts
```diff
@@ -0,0 +1,129 @@
...
+  <label_name>${name}</label_name>
+  <instructions_for_adding_label>${instructions || "No instructions provided"}</instructions_for_adding_label>
+</label>
+
+### Message Content
+<email>
+${stringifyEmail(email, 1000)}
+</email>`;
+
+  logger.trace("Input", { system, prompt });
```
elie222: think it's already traced by `createGenerateObject`

elie222:
Remove WIP when done

apps/web/utils/ai/label-analysis/analyze-label-removal.ts
```diff
@@ -0,0 +1,122 @@
...
+    .describe(
+      "Specific value for the pattern (e.g., email address, domain, subject keyword)",
+    ),
+  exclude: z
+    .boolean()
+    .optional()
+    .describe(
+      "Whether this pattern should be excluded (true) or just not included (false)",
+    ),
+});
```
elie222: This is hardly explained to the AI.
As a reader of this prompt I find it hard to understand how it's used, and I assume it will be the same for the AI.

The prompt as a whole feels unclear.
What even are these patterns it's handling? Define the framework.

Something that can help the AI is sample output. There's an example here:
`apps/web/utils/ai/assistant/chat.ts`

apps/web/utils/ai/label-analysis/analyze-label-removal.test.ts
```diff
@@ -0,0 +1,495 @@
+import { describe, expect, test, vi, beforeEach } from "vitest";
+import { aiAnalyzeLabelRemoval } from "@/utils/ai/label-analysis/analyze-label-removal";
+import { GroupItemType } from "@prisma/client";
+
+// Run with: pnpm test-ai analyze-label-removal
+
+vi.mock("server-only", () => ({}));
+
+// Skip tests unless explicitly running AI tests
+const isAiTest = process.env.RUN_AI_TESTS === "true";
```
elie222: put this in the __tests__  folder as this is currently where all ai tests sit

apps/web/utils/ai/label-analysis/analyze-label-removal.ts
```diff
@@ -0,0 +1,128 @@
...
+
+export type LabelRemovalAnalysis = z.infer<typeof schema>;
+
+export async function aiAnalyzeLabelRemoval({
+  matchedRules,
+  email,
+  emailAccount,
+}: {
+  matchedRules: {
```
elie222: An email matches one rule, so why are we passing in an array of matched rules?

apps/web/utils/ai/label-analysis/analyze-label-removal.ts
```diff
@@ -0,0 +1,128 @@
...
+        `</label_${index}>`,
+      ].join("\n"),
+    )
+    .join("\n")}
+</labels>
+
+Content of the email that has the label removed
+<email>
+${stringifyEmail(email, 1000)}
+</email>`;
```
elie222: Something along these lines would be more clear:

This is the email:
EMAIL

This is the rule we applied to it:
RULE

The user removed the label LABEL.

Decide if we should learn from this for the future.

apps/web/utils/ai/label-analysis/analyze-label-removal.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  const modelOptions = getModel(emailAccount.user, "economy");
```
elie222: We don't need economy here. User's won't be removing emails that often

apps/web/utils/ai/label-analysis/analyze-label-removal.ts
```diff
@@ -0,0 +1,128 @@
...
+  emailAccount: EmailAccountWithAI;
+}): Promise<LabelRemovalAnalysis> {
+  const system = `You are an email expert who manages learned patterns for a user's inbox.  
+You cannot act directly on the inbox; you can only propose learned patterns or suggest no action.  
+Your goal is to help the user manage their emails and labels more effectively and to analyze label removals to identify potential learned patterns.
+
+What are Learned Patterns?
+- Automatically discovered email patterns that consistently trigger the same action.  
+- They tie directly to rules:
+  - Include → always apply the rule
```
elie222: Why would we be doing "Include" pattern here if the user has just removed a label from an email?

It's not given other potential rules so when would it ever want to do this?

I think what we'd want to do:
1. if the rule matched because it matched a learned pattern, ask the AI if the learned pattern should be removed or converted to an EXCLUDE

2. If the rule matched because of an AI instructions match, then the AI should decide if we want an EXCLUDE rule.

3. If the rule matched because it was a static condition match, don't do anything. This is expected behaviour. If the user wants to change this they can adjust their rules.


1 and 2 can be specific focused prompts to the task at hand.

Overall it still seems there's a lot of irrelevant noise in the prompt. Words for the sake of it. Rather than focusing on the actual task.

The prompt should ideally be something along these lines:

We manage this user's inbox. We have a given set of rules and we choose the rule that best matches a given email.
Our AI system matched this email EMAIL to this rule RULE.
The user then removed the label LABEL. Should we adjust our learned patterns?
(And explain what learned patterns are)
And explain we don't always need to learn something, and that it should provide the reasoning.

PR #703, opened 2025-08-20
fix: Clean refresh token only if "invalid_grant"

edulelis:
## Summary by CodeRabbit

* **Bug Fixes**
  * Always prompt Google consent during account connection.
  * Improved Gmail permission checks: validate scopes before/after refresh and attempt refresh when possible.
  * Clearer messaging when access expires, advising reconnection and listing missing permissions.
  * Avoid unconditional deletion of credentials; only clean up stale tokens when refresh confirms expiration.

### DIFF
--- a/apps/web/utils/gmail/permissions.ts
+++ b/apps/web/utils/gmail/permissions.ts
@@ -78,7 +78,7 @@ export async function handleGmailPermissionsCheck({
   emailAccountId,
 }: {
   accessToken: string;
-  refreshToken: string;
+  refreshToken: string | null | undefined;
   emailAccountId: string;
 }) {
   const permissionsBeforeRefresh = await checkGmailPermissions({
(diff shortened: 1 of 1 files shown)


apps/web/utils/gmail/permissions.ts
```diff
@@ -84,7 +84,15 @@ export async function handleGmailPermissionsCheck({
   const { hasAllPermissions, error, missingScopes } =
     await checkGmailPermissions({ accessToken, emailAccountId });
 
-  if (error === "invalid_token") {
+  if (
+    error &&
+    [
+      "invalid_token",
```
edulelis, line 92: Increased the error list for attempting to acquire a new access token.

apps/web/utils/gmail/permissions.ts
```diff
@@ -98,37 +106,48 @@ export async function handleGmailPermissionsCheck({
...
-          accessToken: newAccessToken,
-          emailAccountId,
-        });
+        const { hasAllPermissions, error, missingScopes } =
+          await checkGmailPermissions({
+            accessToken: newAccessToken,
+            emailAccountId,
+          });
+
+        if (error === "invalid_grant") {
```
edulelis: Only clearing refresh token if we get invalid_grant

apps/web/utils/gmail/permissions.ts
```diff
@@ -98,37 +106,48 @@ export async function handleGmailPermissionsCheck({
...
+          logger.info("Cleaning up invalid Gmail tokens", { emailAccountId });
+          const emailAccount = await prisma.emailAccount.findUnique({
+            where: { id: emailAccountId },
```
elie222, line 122: could use select here but minor

apps/web/utils/gmail/permissions.ts
```diff
@@ -97,39 +107,57 @@ export async function handleGmailPermissionsCheck({
...
-        // getGmailClientWithRefresh, getAccessTokenFromClient will throw if access token is invalid
-        // Refresh failed, fall through to cleanup
+        return {
+          hasAllPermissions: false,
+          error: "Gmail access expired. Please reconnect your account.",
+          missingScopes: permissionsBeforeRefresh.missingScopes,
+        };
       }
+    } else {
+      logger.warn("Got no refresh token to attempt refresh", {
```
edulelis, line 156: This should never happen, but added a log here to observe.

apps/web/app/api/google/linking/auth-url/route.ts
```diff
@@ -16,6 +16,7 @@ const getAuthUrl = ({ userId }: { userId: string }) => {
   const url = googleAuth.generateAuthUrl({
     access_type: "offline",
     scope: [...new Set([...SCOPES, "openid", "email"])].join(" "),
+    prompt: "consent",
```
edulelis, line 19: Ideally, we want the user's consent due to the added scopes.

edulelis:
@coderabbitai review

apps/web/app/api/google/linking/auth-url/route.ts
elie222: I don't think it's needed. consent just means if they need to reconsent. If they hadn't given consent they'd see the consent screen anyway. But it's fine to keep this in anyway. It means the user is forced to give consent again even if they already gave consent.

PR #762, opened 2025-09-09
WIP: Org admin permissions

edulelis:


apps/web/app/api/organizations/members/executed-rules-count/route.ts
```diff
@@ -0,0 +1,58 @@
...
+    SELECT 
+      m."userId" as user_id,
+      COUNT(er.id) as executed_rules_count
+    FROM "Member" m
+    JOIN "User" u ON m."userId" = u.id
+    JOIN "EmailAccount" ea ON u.id = ea."userId" AND ea.email = u.email
+    LEFT JOIN "ExecutedRule" er ON ea.id = er."emailAccountId"
+    WHERE m."organizationId" = ${organizationId}
+    GROUP BY m."userId"
+  `;
```
elie222: is this the only way to do this?
in general i prefer not to use raw sql queries, unless it can't be done with prisma / clear perf benefits to writing a raw query.

apps/web/app/api/organizations/members/executed-rules-count/route.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  // Transform the result to match expected format
```
elie222: can drop these ai comments

apps/web/hooks/useOrgAccess.ts
```diff
@@ -0,0 +1,53 @@
...
+    async (url) => {
+      const response = await fetchWithAccount({ url, emailAccountId });
+      if (!response.ok) {
+        const errorText = await response.text();
+        throw new Error(
+          `Failed to fetch email account: ${response.status} ${errorText}`,
+        );
+      }
+      return response.json();
+    },
```
elie222: what's the reason `async (url) => {` func was needed? the code doesnt work otherwise?

apps/web/hooks/useOrgAccess.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  if (!session?.user?.email || !emailAccount) {
+    return {
+      isLoading: true,
+      isAccountOwner: true,
+      accountInfo: null,
+    };
+  }
+
+  const isAccountOwner = emailAccount.email === session.user.email;
```
elie222: `session.user.email` is the `User` email address.

user can have 5 email accounts under it. but all own the account. they can all log in to the user. in general session.user.email isn't used for much anymore.

apps/web/hooks/useOrgAccess.ts
```diff
@@ -0,0 +1,53 @@
...
+      email: emailAccount.email,
+      name: emailAccount.name,
+      image: emailAccount.image,
+      provider: emailAccount.account?.provider,
+    },
+  };
+}
+
+export function isViewingDifferentAccount(
+  currentUserEmail: string,
```
elie222: note that a user can have 5 emails under it

apps/web/hooks/useOwnEmailAccount.ts
```diff
@@ -0,0 +1,18 @@
+import useSWR from "swr";
+import type { GetEmailAccountsResponse } from "@/app/api/user/email-accounts/route";
+
+export function useOwnEmailAccount() {
+  const { data, isLoading } = useSWR<GetEmailAccountsResponse>(
+    "/api/user/email-accounts",
+  );
```
elie222: you can use `useAccounts()` which already exists and calls this endpoint.

and then you can replace the entire hook. the only thing it adds is which could just be a simple util function:

```
const ownEmailAccount = data?.emailAccounts?.find(
    (account) => account.isPrimary,
  );
```

apps/web/utils/actions/safe-action.ts
```diff
@@ -7,14 +7,22 @@ import prisma from "@/utils/prisma";
...
 const baseClient = createSafeActionClient({
   defineMetadataSchema() {
-    return z.object({ name: z.string() });
+    return z.object({
+      name: z.string(),
+      options: z
+        .object({
+          allowAdmins: z.boolean().nullish(),
+        })
+        .nullish(),
```
elie222: 1. not sure it needs to be nested in options
2. it's not used right now, so maybe just remove it? or comment it out?

the reason for 2 is that it's room for bugs / security holes. so better we add it when it's in actual use and we can better check it

apps/web/utils/actions/safe-action.ts
```diff
@@ -68,8 +76,32 @@ export const actionClient = baseClient
...
+                emailAccount: targetEmailAccount,
+                provider: targetEmailAccount.account.provider,
+              },
+            });
+          });
+        }
+      }
+
       throw new SafeError("Unauthorized");
+    }
```
elie222: can remove this for now, as it's not in use

apps/web/utils/middleware.ts
```diff
@@ -39,17 +40,25 @@ export interface RequestWithEmailProvider extends RequestWithEmailAccount {
   emailProvider: EmailProvider;
 }
 
+export interface MiddlewareOptions {
+  allowAdmins?: boolean;
```
elie222: change this to `allowOrgAdmin` or `s`

so we don't get confused in the future between org admin and site admin, which is how the word is used in other places like for the `/admin` screen and permissions

apps/web/utils/middleware.ts
```diff
@@ -39,17 +40,25 @@ export interface RequestWithEmailProvider extends RequestWithEmailAccount {
...
+}
+
 // Higher-order middleware factory that handles common error logic
 function withMiddleware<T extends NextRequest>(
   handler: NextHandler<T>,
-  middleware?: (req: NextRequest) => Promise<T | Response>,
+  middleware?: (
+    req: NextRequest,
+    options?: MiddlewareOptions,
```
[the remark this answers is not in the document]
elie222: if this can be ignored just resolve it

apps/web/utils/middleware.ts
```diff
@@ -134,6 +143,7 @@ function withMiddleware<T extends NextRequest>(
 
 async function authMiddleware(
   req: NextRequest,
+  options?: MiddlewareOptions,
```
elie222: remove if not used here?

apps/web/utils/middleware.ts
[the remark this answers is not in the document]
elie222: ah, seems like the note is correct as it's not used here? if so, then just remove it

apps/web/utils/middleware.ts
```diff
@@ -187,6 +216,7 @@ async function emailAccountMiddleware(
 
 async function emailProviderMiddleware(
   req: NextRequest,
+  options?: MiddlewareOptions,
```
elie222: remove?

apps/web/utils/organizations/access.ts
```diff
@@ -0,0 +1,51 @@
+import prisma from "@/utils/prisma";
+import { hasOrganizationAdminRole } from "./roles";
+
+export async function checkOrgEmailAccountAccess(
+  userId: string,
+  emailAccountId: string,
```
elie222: rename this to make it clear this is the target email account

apps/web/utils/organizations/access.ts
```diff
@@ -0,0 +1,51 @@
...
+          members: {
+            where: {
+              organizationId: userMembership.organizationId,
+            },
+            select: {
+              id: true,
+            },
+          },
+        },
+      },
```
elie222: do we always need to fetch the memberid? do we use it at all?
note, if there's a specific spot you want to fetch memberid, you could always just fetch it where needed. when it happens in the middleware it happens everytime

elie222: ah. i see what you're doing. i think this is the wrong way to do it.

it would be this instead:

```ts
const targetEmailAccount = await prisma.emailAccount.findFirst({
    where: { id: emailAccountId, user: member: organizationId },
```

And then you don't need the select at all. Also the code below would change. It would just return null

apps/web/hooks/useOrgAccess.ts
edulelis: I refactored this to use a different function `useOrgSWR`, but the key difference of this and others is that this needs the email account from the URL param and not from the logged-in user.
This one is: an admin fetching email-account info from a member.

apps/web/app/api/organizations/members/executed-rules-count/route.ts
edulelis: refactored to use a Prisma query.

apps/web/utils/organizations/access.ts
edulelis: refactored to
```
 const targetEmailAccount = await prisma.emailAccount.findFirst({
    where: { 
      id: targetEmailAccountId,
      user: {
        members: {
          some: {
            organizationId: callerMember.organizationId,
          },
        },
      },
    },
  });
  ```

PR #822, opened 2025-09-27
Phase 1: Remove premium validation checks - unlock all features

parthmodi152:
- Modified premium validation functions to always return true
- All users now have full access to AI features
- Unlimited unsubscribe operations enabled
- Knowledge base restrictions removed
- Maximum tier (BUSINESS_PLUS_MONTHLY) granted to all users
- Simplified server-side premium checks
- Fixed linting issues


elie222:
You can just upgrade yourself in admin area btw

PR #842, opened 2025-10-10
Add links to digested emails

edulelis:
## Summary by CodeRabbit

- New Features
  - Digest emails now include per-item "View in inbox" links when available; Outlook web links are honored and Gmail links include encoded address context. Digest items may carry an optional emailUrl.

- Refactor
  - UI components now accept a single message object prop instead of many separate fields; URL helpers unified with safer provider fallbacks and consistent id/thread handling.

- Tests
  - Added unit tests covering URL generation, provider behaviors, encoding, and fallbacks.


apps/web/app/(app)/[emailAccountId]/assistant/ExecutedRulesTable.tsx
```diff
@@ -186,7 +175,7 @@ function OpenInGmailButton({
 
   return (
     <Link
-      href={getEmailUrlForMessage(messageId, threadId, userEmail, provider)}
```
Comment (edulelis) [inline apps/web/app/(app)/[emailAccountId]/assistant/ExecutedRulesTable.tsx]: Modified this function signature to pass the entire message instead of just messageId and threadId.
Reason was that for Outlook, the link will be sourced from the message.webLink and gmail will use the message.id

Issue #401, opened 2025-04-13
Docker-compose.yml requires compose v2 (Readme issue)  /  .env path mismatch (environment specific?) in docker-compose


milankubin:
https://github.com/elie222/inbox-zero/blob/1f3bb2b3f486ffa393430c4547d4b2c0fee6e466/docker-compose.yml#L45

Running on Debian, docker-compose.yml does not work on v1.x versions of compose (Default for any Debian stable release):
``` bash
└╼milan$docker-compose up -d 
ERROR: The Compose file './../../docker-compose.yml' is invalid because:
'name' does not match any of the regexes: '^x-'
```

Had to install docker-compose v2 from experimental branch, then:

``` bash
docker-compose up -a
```
now becomes 
``` bash
docker compose up -a
```

produces the following error
``` bash
WARN[0000] The "UPSTASH_REDIS_TOKEN" variable is not set. Defaulting to a blank string. 
WARN[0000] The "UPSTASH_REDIS_TOKEN" variable is not set. Defaulting to a blank string. 
env file <omitted_local_path>/inbox-zero/.env not found: stat <omitted_local_path>/inbox-zero/.env: no such file or directory
```
Updating aforementioned line in docker-compose to use './apps/web/.env'  fixed the issue.

vishalmakwana111:
Hi @elie222, I've opened PR #411 to address the `env_file` path confusion mentioned in this issue. It changes the `docker-compose.yml` file to point the `web` service's `env_file` directly to `apps/web/.env`, simplifying the setup for Docker users.

eethann:
I just ran into the UPSTASH_REDIS_TOKEN issue when installing locally. The specific issue, as I understand it, is that the docker-compose.yml file looks to the "root" environment for variables, not the `apps/web/.env` environment. I think this can be 

I addressed it with a quick short-term fix:

```
export UPSTASH_REDIS_TOKEN=token-string-here
```

And then I created a .env at the root level of the project with the `UPSTASH_REDIS_TOKEN` env variable.

That seems to work on a restart of the docker container, but I'm not convinced it's not relying on the old env var.

KarolKozlowski:
The file `apps/web/.env` does NOT exist in the repository which breaks the compose file - this should exist and contain reasonable defaults. Currently it points to a non existing file and breaks the config:

```
❯ docker compose config
WARN[0000] The "UPSTASH_REDIS_TOKEN" variable is not set. Defaulting to a blank string.
[...]

PR #954, opened 2025-11-12
fix: Point env_file to an existing file (issue #401)

KarolKozlowski:
**Problem:**
I want to be able to use the source repository without any modifications in my deployment, so I can easily keep it up to date (for example as a git submodule). Currently the compose file is broken as it points to a non-existing file.

**Solution:**
Modify `docker-compose.yml`, so that the `env_file` variable points to an existing file with reasonable defaults. This can be overridden with any of the standard docker compose practices. Updated the `docker.md` to reflect this change and since the override file is `.env` no explicit `--env-file` is necessary. 

**Note:**
Alternatively a `./apps/web/.env` could be added to `.gitignore` which would allow to create an env file that would not appear as the change in the repository, however having a consistent compose file makes more sense IMO.

## Summary by CodeRabbit

* **Documentation**
  * Simplified Docker setup instructions and updated startup command to use standard compose invocation.
  * Added guidance on overriding volumes for advanced setups and multi-file compose usage.

* **Chores**
  * Streamlined environment file handling in Docker configuration to simplify deployment and local setup.

### DIFF
--- a/docs/hosting/docker.md
+++ b/docs/hosting/docker.md
@@ -36,7 +36,7 @@ and edit with your production settings.
 
 For detailed configuration instructions including all required environment variables, OAuth setup, and LLM configuration, see the [main README.md configuration section](../../README.md#updating-env-file-secrets).
 
-If you need to adjust volumes ort other parameters you can override them with `docker-compose.override.yml`, please refer to [official documentation](https://docs.docker.com/compose/how-tos/multiple-compose-files/) on how to manage multiple compose files.
+If you need to adjust volumes or other parameters you can override them with `docker-compose.override.yml`, please refer to [official documentation](https://docs.docker.com/compose/how-tos/multiple-compose-files/) on how to manage multiple compose files.
 
 ### 4. Build and Deploy
 


elie222:
Hey, you need to copy the .env.example to a .env file for it to work. What's the reason you don't want to do that?

We don't want people using unsecure hardcoded .env.example. It should be created by the user and have their secrets in it

KarolKozlowski:
Here is the thing - the example compose file should be a working example - currently it isn't. it should be possible to reference your compose file without ANY modifications and be able to override it with one of the standard ways described [here](https://docs.docker.com/compose/how-tos/multiple-compose-files/), currently the file points to an non-existing env file which CANNOT be overridden in any way according to the official compose documentation.

```
❯ docker compose config
env file /home/karol.kozlowski/wa/inbox-zero/apps/web/.env not found: stat /home/karol.kozlowski/wa/inbox-zero/apps/web/.env: no such file or directory
```

This cannot be overridden, the compose file in the current state is BROKEN:

```
❯ docker compose --env-file .env-my-local-overrides config
env file /home/karol.kozlowski/wa/inbox-zero/apps/web/.env not found: stat /home/karol.kozlowski/wa/inbox-zero/apps/web/.env: no such file or directory
```

The hardcoded `env_file` could either be pointing to a reasonable global defaults (secrets should be omitted) or be removed completely to provide a properly formatted compose file. 

Also, currently the deployment instructions have the following line `cp apps/web/.env.example apps/web/.env` how is this more 'secure'?

elie222:
I don't understand what you're saying. How is the file supposed to run with an example env file?

PR #988, opened 2025-11-19
Add LM Studio Provider

### LINKED ISSUE #507: FEATURE REQUEST: LM-Studio API
would you be willing to support one of the lm-studio api's?  It appears that lm-studio has their own beta rest api (subject to change), but they also have an openai compatibility api.  Given you already support openai, i assume it may not be a significant amount of work to interface with their openai api.

[LM-Studio Rest API](https://lmstudio.ai/docs/app/api/endpoints/rest)

[LM-Studio OpenAI Compatible API](https://lmstudio.ai/docs/app/api/endpoints/openai)

The reason I want to use LM-Studio is because ollama does not have support for MLX (used in apple silicon), but lm-studio has full support.  LM-Studio performs significantly better than ollama on macs as result.

ppranay20:
Closes #507 

## Description
Integrates **LM Studio** as a supported AI provider using the OpenAI-compatible SDK. 

## Changes
- Added `LM_STUDIO` as a provider in `model.ts`.
- Added tests in `model.ts` for LM_STUDIO.
- Updated `env.example` to include NEXT_PUBLIC_LM_STUDIO_MODEL and LM_STUDIO_BASE_URL.
- Patched `chat` and `economy` model resolvers to ensure they return correctly when a local provider is selected.

## Summary by CodeRabbit

* **New Features**
  * Added LM Studio as a supported language-model provider and UI/runtime options to select its model.
  * Integrated OpenAI-compatible connectivity for LM Studio backends.

* **Chores**
  * Added an OpenAI-compatible SDK dependency.

* **Tests**
  * Added tests validating LM Studio configuration and integration.

### DIFF
--- a/apps/web/utils/llms/model.ts
+++ b/apps/web/utils/llms/model.ts
@@ -144,12 +144,20 @@ function selectModel(
       const modelName = aiModel || Model.LM_STUDIO;
 
       if (!modelName) {
-        throw new Error("For LM_STUDIO, 'LLM_MODEL' must be set.");
+        throw new Error(
+          "LM Studio model not configured. Please set NEXT_PUBLIC_LM_STUDIO_MODEL environment variable.",
+        );
+      }
+
+      if (!env.LM_STUDIO_BASE_URL) {
+        throw new Error(
+          "LM Studio base URL not configured. Please set LM_STUDIO_BASE_URL environment variable.",
+        );
       }
 
       const lmstudio = createOpenAICompatible({
         name: Provider.LM_STUDIO,
-        baseURL: env.LM_STUDIO_BASE_URL!,
+        baseURL: env.LM_STUDIO_BASE_URL,
         supportsStructuredOutputs: true,
       });
 


apps/web/utils/llms/model.ts
```diff
@@ -211,7 +240,17 @@ function createOpenRouterProviderOptions(
...
   if (env.ECONOMY_LLM_PROVIDER && env.ECONOMY_LLM_MODEL) {
+    const isLMStudio = env.ECONOMY_LLM_PROVIDER === Provider.LM_STUDIO;
+    if (isLMStudio) {
+      return selectModel({
+        aiProvider: Provider.LM_STUDIO,
+        aiModel: env.ECONOMY_LLM_MODEL,
+        aiApiKey: null,
+      });
+    }
+
```
elie222, line 251: Seems inconsistent? And would ignore the user settings if those are set?

Same in the other plcae you did this

ppranay20: If I don't add this then it will direct to default function (because LM studio does not have API key it provides URL). The default function selects the model which is set to default. So, if economy is set to LM studio it will revert to default model. 

I can also add a check that for LM studio URL check like 
`if (!apiKey && !url) {
      logger.warn("Economy LLM provider configured but API key not found", {
        provider: env.ECONOMY_LLM_PROVIDER,
      });
      return selectDefaultModel(userAi);
    }`
    
        
This will also work.

justinbadal:
Hello, is this still being worked on?

elie222:
> Hello, is this still being worked on?

no. please create a new pr with a fix.

others want this too. if you make a new PR please share your prompt in the contribution too

PR #995, opened 2025-11-20
Charts redesign

jshwrnr:
## Summary by CodeRabbit

* **New Features**
  * Added tabbed bar chart visualizations for enhanced data exploration in analytics.
  * Improved date range selection with relative date options and semantic labels.

* **Enhancements**
  * Reorganized analytics dashboard controls for clearer interface layout.
  * Enhanced progress indicators with tooltip information display.
  * Updated filtering options with improved visual indicators.
  * Refined button styling and dropdown interactions.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


elie222:
Amazing update. Looks so much better than what we had before. Some notes:

1. Minor: empty states (e.g. send is empty)

CleanShot 2025-11-20 at 15 19 32

CleanShot 2025-11-20 at 15 19 28

2. Major:
Something is wrong with group / history. e.g oct 19 shows twice. And also this isn't grouped by day:

CleanShot 2025-11-20 at 15 23 31

3. Should we have "filter" on the left and "group by" next to it (as filter is the primary action)

This is what Stripe does:

CleanShot 2025-11-20 at 16 09 47

4. Bulk unsubscriber page uses a different date filter without the sidebar? Would be nice to use same component for both:

CleanShot 2025-11-20 at 15 26 45

5. Deep Clean - always showing? I'm taking a look at this so don't worry about it.

apps/web/app/(app)/[emailAccountId]/stats/mock-data.ts
```diff
@@ -0,0 +1,26 @@
+import type { EmailActionStatsResponse } from "@/app/api/user/stats/email-actions/route";
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/stats/mock-data.ts]: Delete file

apps/web/app/(app)/[emailAccountId]/stats/RuleStatsChart.tsx
```diff
@@ -76,7 +80,18 @@ export function RuleStatsChart({ dateRange, title }: RuleStatsChartProps) {
       ),
     };
 
-    return { pieChartData: pieData, chartConfig: config };
+    const barConfig: ChartConfig = {
+      executed: {
+        label: "Executed Rules",
+        color: "#006EFF80",
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/stats/RuleStatsChart.tsx]: we have a lot of magic strings like this in the codebase. also for the landing page stuff. ideally we put it into tailwind config. in the shadcn code examples it seems like you can do this.

but if we do need to use a variable, would still be good to put them into a single file and name the colors

apps/web/app/(app)/[emailAccountId]/stats/Stats.tsx
```diff
@@ -71,63 +66,53 @@ export function Stats() {
...
+          <List
+            value={
+              selectOptions.find((option) => option.label === dateDropdown)
+                ?.value
+            }
+            items={selectOptions}
+            className="min-w-32"
+            onSelect={({ label, value }) => {
+              onSetDateDropdown({ label, value });
+              setDateRange({
```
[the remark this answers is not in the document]
Comment (elie222) [reply apps/web/app/(app)/[emailAccountId]/stats/Stats.tsx]: @jshwrnr check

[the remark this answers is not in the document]
Comment (elie222) [reply apps/web/app/(app)/[emailAccountId]/stats/Stats.tsx]: @jshwrnr

apps/web/components/ui/button.tsx
```diff
@@ -30,7 +30,7 @@ const buttonVariants = cva(
       },
       size: {
         default: "h-10 px-4 py-2",
-        xs: "h-6 rounded-sm px-1.5 text-xs",
+        xs: "h-7 rounded-md px-2 text-xs",
```
elie222: This changes the entire site. If you want a new button variant add it instead of updating styling everywhere

elie222: Is this for the watch button? if so we may not even need this change at all

apps/web/components/ui/progress.tsx
```diff
@@ -4,21 +4,24 @@ import * as React from "react";
...
     className={cn(
-      "relative h-4 w-full overflow-hidden rounded-full bg-secondary",
+      "relative h-2 w-full overflow-hidden rounded-full bg-secondary",
       className,
     )}
     {...props}
   >
     <ProgressPrimitive.Indicator
-      className="h-full w-full flex-1 bg-primary transition-all"
+      className={cx("h-full w-full flex-1 transition-all", color)}
```
elie222: instead of passing color as an individual prop, you can pass className (or something like innerClassName as it would override the above)

apps/web/components/ui/progress.tsx
```diff
@@ -4,21 +4,24 @@ import * as React from "react";
...
->(({ className, value, ...props }, ref) => (
+  React.ComponentPropsWithoutRef<typeof ProgressPrimitive.Root> & {
+    color?: string;
+  }
+>(({ className, value, color = "bg-blue-500", ...props }, ref) => (
   <ProgressPrimitive.Root
     ref={ref}
     className={cn(
-      "relative h-4 w-full overflow-hidden rounded-full bg-secondary",
+      "relative h-2 w-full overflow-hidden rounded-full bg-secondary",
```
elie222: I'd avoid changing the shadcn stuff unless the intention really is for this to be changed everywhere. you can pass className="h-2" without changing this at all

apps/web/components/ui/calendar.tsx
```diff
@@ -3,26 +3,34 @@
...
 import { cn } from "@/utils";
 import { buttonVariants } from "@/components/ui/button";
+import { cx } from "class-variance-authority";
 
-export type CalendarProps = React.ComponentProps<typeof DayPicker>;
+export type CalendarProps = React.ComponentProps<typeof DayPicker> & {
+  rightContent?: React.ReactNode;
+};
 
 function Calendar({
```
elie222, line 13: lots of changes to Calendar. in general im not a fan of changing the shadcn components. in this case could you have created a new component that looks something like this:

```tsx
<div><Calendar/><Sidebar/></div>
```

If that wasn't possible here then what you did is fine, but as a general rule I'd try leave the shadcn stuff as is unless we really intend for everything across the app to be changed

apps/web/components/StatsCards.tsx
elie222, line 1: StatsCards is still in use. You need to be more careful with the sweeping changes. If you delete a file you need to check all places it's used.

Similar to the shadcn component changes. The changes impact lots of places.

Image

apps/web/components/TabSelect.tsx
```diff
@@ -15,7 +15,7 @@ import Link from "next/link";
 import { type Dispatch, type SetStateAction, useId } from "react";
 import { ArrowUpRight } from "lucide-react";
 
-const tabSelectButtonVariants = cva("p-4 transition-colors duration-75", {
+const tabSelectButtonVariants = cva("p-3 transition-colors duration-75", {
```
elie222: again we have changes to the entire app. is this really intended? have you checked every place we use it?

apps/web/components/List.tsx
```diff
@@ -0,0 +1,39 @@
+import { Button } from "@/components/ui/button";
```
elie222, line 1: Why adding a `components/common` folder?
Stick with the existing pattern and put the components in `components`

apps/web/components/DatePickerWithRange.tsx
```diff
@@ -13,48 +12,100 @@ import {
...
+      ? differenceInDays(dateRange.to, dateRange.from)
+      : 0;
+
+  const relativeDateLabel =
+    days === 1
+      ? "Last day"
+      : days === 7
+        ? "Last week"
+        : days === 30
+          ? "Last month"
```
elie222: This is really ugly. You can probably have a cleaner result with a function. Rough idea:

```tsx
getLabel = () => {
if (days === 1) return "Last day"
if (days === 7) return "Last week"
}
```

Or a switch statement in that func if you prefer.

apps/web/components/ui/calendar.tsx
```diff
@@ -3,26 +3,34 @@
...
   className,
   classNames,
   showOutsideDays = true,
+  rightContent,
   ...props
 }: CalendarProps) {
+  const padding = "p-3";
+  const borderLeft = "border-l border-gray-200";
```
elie222: Feels ugly. I'd just inline the classes where they're used.

So someone can just read caption_end and see the styling for it.

From LLM:

Why it feels wrong:
Pollutes the Render Scope: Defining styling constants like const padding = "p-3" inside the component body adds noise. Every time you read the component logic, you have to parse these style definitions that aren't actually logic.
Unnecessary Indirection: It forces you to jump your eyes up and down to see what padding or borderLeft actually equals. For utility classes (Tailwind), the convention is almost always to keep them inline so you can see the styles right where they are applied.
Non-Standard: As you noted, this isn't standard practice. In React + Tailwind, we typically either inline the strings or, for complex conditional logic, use cn() / clsx. Creating variables for simple static strings like "p-3" is over-engineering.
The Better Approach:
The better approach (which I applied in the previous step) is to inline the strings directly.
Instead of:
const padding = "p-3";// ...caption_start: padding,
Just do:
caption_start: "p-3",caption_end: "p-3 border-l border-gray-200",
This makes the code declarative and easier to scan. I removed the variables and the cx import since they weren't adding value

apps/web/components/ui/calendar.tsx
```diff
@@ -3,26 +3,34 @@
...
     <DayPicker
       showOutsideDays={showOutsideDays}
-      className={cn("p-3", className)}
+      className={cn("", className)}
       classNames={{
         months: "flex flex-col sm:flex-row space-y-4 sm:space-x-4 sm:space-y-0",
         month: "space-y-4",
         caption: "flex justify-center pt-1 relative items-center",
+        caption_start: padding,
+        caption_end: cx(padding, borderLeft),
```
elie222: Don't think we need cx. We have cn.

But also, won't need it once we inline the styles

apps/web/components/ui/calendar.tsx
```diff
@@ -3,26 +3,34 @@
 import type * as React from "react";
 import { ChevronLeft, ChevronRight } from "lucide-react";
 import { DayPicker } from "react-day-picker";
-
 import { cn } from "@/utils";
 import { buttonVariants } from "@/components/ui/button";
+import { cx } from "class-variance-authority";
```
elie222: we use `cn`. But this will get removed anyway when you make the change below

apps/web/components/charts/DomainIcon.tsx
```diff
@@ -0,0 +1,75 @@
...
+        "rounded-full size-5 z-10 bg-gradient-to-r",
+        gradients[gradientIndex],
+      )}
+    />
+  );
+}
+
+interface DomainIconProps {
+  domain: string;
+}
```
elie222, line 52: in general fine to just inline it:

```tsx
export function DomainIcon({ domain }:{ domain: string }) {
```

but not critical to change. Cursor will often break into its own thing anyway

jshwrnr: Alright 👍🏻

apps/web/components/ui/button.tsx
jshwrnr: I made the change for the "Load more" button, but in the end am not using xs for it. I'll undo this change.

apps/web/components/ui/calendar.tsx
jshwrnr: There was no good way to do it as I needed to inject the sidebar into the `Months` component of react-day-picker which the shadcn calendar is using. I'll avoid making further changes to shad/cn components but this one was necessary in this case.

apps/web/components/StatsCards.tsx
jshwrnr: Yeh my bad, will be more careful with that. I brought back the StatsCards component for the usage page.

apps/web/app/(app)/[emailAccountId]/stats/RuleStatsChart.tsx
Comment (jshwrnr) [reply apps/web/app/(app)/[emailAccountId]/stats/RuleStatsChart.tsx]: cleaned up most of it

jshwrnr:
> Amazing update. Looks so much better than what we had before. Some notes:
> 
> 1. Minor: empty states (e.g. send is empty)
> 
> CleanShot 2025-11-20 at 15 19 32 CleanShot 2025-11-20 at 15 19 28
> 2. Major:
>    Something is wrong with group / history. e.g oct 19 shows twice. And also this isn't grouped by day:
> 
> CleanShot 2025-11-20 at 15 23 31
> 3. Should we have "filter" on the left and "group by" next to it (as filter is the primary action)
> 
> This is what Stripe does:
> 
> CleanShot 2025-11-20 at 16 09 47
> 4. Bulk unsubscriber page uses a different date filter without the sidebar? Would be nice to use same component for both:
> 
> CleanShot 2025-11-20 at 15 26 45
> 5. Deep Clean - always showing? I'm taking a look at this so don't worry about it.

I added a Notion task to discuss number 3: https://www.notion.so/elietech/Should-we-have-filter-on-the-left-and-group-by-next-to-it-as-filter-is-the-primary-action-2b4f403a1421807fad79ebdd140c621e?source=copy_link

PR #1148, opened 2025-12-31
security: implement 4 security hardening fixes

rsnodgrass:
## Summary

This PR implements 4 security hardening measures:

### 1. Encrypt User AI API Keys and Webhook Secrets at Rest
- Added AES-256-GCM encryption for `User.aiApiKey` and `User.webhookSecret`
- Uses `enc:` prefix to reliably detect encrypted values
- Graceful degradation: plaintext values work but log warnings
- Migration script: `apps/web/scripts/encrypt-existing-user-secrets.ts`

### 2. Remove Secrets from `/api/user/me` Response
- `/api/user/me` now returns `hasAiApiKey` and `hasWebhookSecret` booleans
- New dedicated `/api/user/secrets` endpoint for actual secret values
- Frontend components updated to use the new secrets endpoint
- Protected with `withAuth` middleware

### 3. Warn When Webhook Verification Tokens Not Set
- Added warning logs when `GOOGLE_PUBSUB_VERIFICATION_TOKEN` not set
- Added warning logs when `MICROSOFT_WEBHOOK_CLIENT_STATE` not set
- Non-blocking: allows development without these tokens

### 4. Prompt Injection Mitigation
- Added `@presidio-dev/hai-guardrails` library for heuristic detection
- Fast (<1ms) detection with no API calls required
- Prompt hardening instructions added to AI system prompts
- Content sanitization: XML tag escaping, control character removal
- Defense-in-depth approach: detect, sanitize, and harden

## Test Plan

- [ ] Verify encryption works for new API keys saved via settings
- [ ] Verify existing plaintext keys still work (with warning in logs)
- [ ] Run migration script to encrypt existing values
- [ ] Verify `/api/user/me` no longer exposes secrets
- [ ] Verify settings pages load secrets correctly via `/api/user/secrets`
- [ ] Verify webhook handlers log warnings when tokens not configured
- [ ] Verify AI rule selection works with prompt hardening

## Release Notes

* **New Features**
  * Rules can now be imported from and exported to JSON files for backup and sharing.

* **Improvements**
  * Enhanced security with encryption for API keys and webhook secrets.
  * Improved AI rule selection with additional safety guardrails against prompt injection.
  * Better error handling and messaging for webhook verification configuration.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -50,7 +50,6 @@
     "@next/third-parties": "16.1.1",
     "@openrouter/ai-sdk-provider": "1.4.0",
     "@portabletext/react": "4.0.3",
-    "@presidio-dev/hai-guardrails": "^1.11.1",
     "@prisma/adapter-pg": "7.1.0",
     "@prisma/client": "6.16.3",
     "@radix-ui/react-alert-dialog": "1.1.15",
--- a/apps/web/utils/ai/guardrails.ts
+++ b/apps/web/utils/ai/guardrails.ts
@@ -1,24 +1,6 @@
-import {
-  injectionGuard,
-  GuardrailsEngine,
-  type GuardrailResult,
-} from "@presidio-dev/hai-guardrails";
 import { createScopedLogger } from "@/utils/logger";
 
 const logger = createScopedLogger("ai-guardrails");
-
-// Create a reusable injection guard with heuristic mode (fast, no API calls)
-const promptInjectionGuard = injectionGuard(
-  { roles: ["user"] },
-  {
-    mode: "heuristic",
-    threshold: 0.7,
-  },
-);
-
-const guardrailsEngine = new GuardrailsEngine({
-  guards: [promptInjectionGuard],
-});
 
 /**
  * Prompt hardening instructions to prepend to system prompts.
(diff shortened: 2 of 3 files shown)


rsnodgrass:
@elie222 this may be something you want to look at ASAP, notably prompt injection

elie222:
> @elie222 this may be something you want to look at ASAP, notably prompt injection

Thanks. It's limited what the AI can do. We expect any content to be able to be sent over email.

Can we choose a different library to this one:
https://www.npmjs.com/package/@presidio-dev/hai-guardrails

It doesn't have much usage and I'm more worried about the package being an attack surface than it helping against injection.

rsnodgrass:
> Can we choose a different library to this one: https://www.npmjs.com/package/@presidio-dev/hai-guardrails
> 
> It doesn't have much usage and I'm more worried about the package being an attack surface than it helping against injection.

Good point. Removed and replaced it with a simple built-in heuristic detector (no dependencies). Now it does the following (but doesn't block potential emails that are suspicious where patterns were unable to be sanitized):

  - Sanitize - Neutralizes dangerous patterns by transforming them
  - Warn - Logs suspicious content for monitoring
  - Harden - Prompt instructions tell AI to treat email as untrusted data

This is much better than the current state, but there is probably room for improvement long term with this.

elie222:
I took over this similarly to other PRs. This has a few different features mixed into one, so it's a little bit hard to merge right now. 

Regarding the prompt hardening, in a lot of cases, I'm not sure it's relevant because the AI is really just labeling emails and drafting replies, and the user can look over both of them in any case before sending. 

There are cases where a user can do something which is riskier, which is automatically send emails with AI-generated content. We have a `getRiskLevel` that determines the risk. I'm wondering if we only need the prompt hardening for those cases. Most users don't have rules like that, and when they do, the UI displays that the rule is risky.

PR #1150, opened 2025-12-31
feat: Recipient-Adaptive Tone - Personalized draft style per recipient

rsnodgrass:
## Summary

@elie222 This is a feature suggestion, feel free to delete or incorporate.

Tracks communication patterns per recipient to adapt draft tone. More formal for executives/external partners, casual for close colleagues.

Example: Draft to CEO uses "Dear Mr. Smith" while draft to teammate uses "Hey!"

## Changes

- **New model**: `RecipientProfile` in Prisma schema
  - Stores detected formality, typical greetings/signoffs per recipient
  - Unique constraint on (emailAccountId, recipientEmail)

- **New file**: `apps/web/utils/ai/knowledge/recipient-profile.ts`
  - `aiAnalyzeRecipientProfile()` - AI analysis of sent emails to a recipient
  - `getRecipientProfile()` - Fetch stored profile from database
  - `saveRecipientProfile()` - Save analysis results
  - `formatRecipientProfileForPrompt()` - Format for AI consumption

- **Modified**: `apps/web/utils/ai/reply/draft-with-knowledge.ts`
  - Added `recipientProfileContext` parameter to use communication style

- **Modified**: `apps/web/utils/reply-tracker/generate-draft.ts`
  - Integrated profile lookup into draft generation pipeline

## How it works

1. When generating a draft, the system looks up the stored RecipientProfile
2. If a profile exists, it includes the detected formality/greeting/signoff style
3. The AI adapts the draft to match how the user typically communicates with that recipient

## Future enhancements (not in this PR)

- Trigger analysis when sending emails (currently profile must be pre-populated)
- Background job to bootstrap profiles from existing sent emails
- UI to view/override recipient profiles

## Test plan

- [ ] Verify RecipientProfile model works correctly
- [ ] Test profile formatting for AI prompts
- [ ] Confirm draft generation includes profile context when available

### DIFF
--- a/apps/web/utils/ai/knowledge/recipient-profile.ts
+++ b/apps/web/utils/ai/knowledge/recipient-profile.ts
@@ -42,7 +42,7 @@ export async function aiAnalyzeRecipientProfile(options: {
   const { sentEmails, recipientEmail, emailAccount } = options;
 
   if (sentEmails.length < 2) {
-    logger.debug("Insufficient emails for recipient profile analysis", {
+    logger.trace("Insufficient emails for recipient profile analysis", {
       recipientEmail,
       emailCount: sentEmails.length,
     });


rsnodgrass:
@elie222 I had some extra Claude Code credits expiring December 31 and wanted to donate them to the project by implementing these features as proof of concepts. Feel free to use, modify, or delete/throw out as you see fit!

**Why Recipient-Adaptive Tone is a great addition:**

People communicate differently with different people - formal with the CEO, casual with close teammates, somewhere in between with clients. Currently, the AI drafts all replies in one style. This feature learns how the user actually communicates with each recipient and adapts accordingly.

Instead of generic "Dear X" or always using "Hi", the draft matches what the user would naturally write: "Hey!" to a close colleague, "Dear Mr. Smith" to an executive, etc. This makes AI-generated drafts feel much more authentic and reduces the editing users need to do.

The implementation stores per-recipient profiles (formality level, typical greeting/signoff) that can be built up over time as the user sends emails. It's a differentiator that competitors like Superhuman don't offer yet.

apps/web/prisma/schema.prisma
```diff
@@ -175,11 +175,38 @@ model EmailAccount {
...
+  communicationFrequency Int? // emails per month to this recipient
+
+  // metadata
+  sampleCount    Int       @default(0) // number of emails analyzed
+  lastAnalyzedAt DateTime?
+  createdAt      DateTime  @default(now())
+  updatedAt      DateTime  @updatedAt
+
+  @@unique([emailAccountId, recipientEmail])
+  @@index([emailAccountId])
```
elie222, line 206: minor, but dont think we need the index with the unique index above. i know llms love to add this though

apps/web/prisma/schema.prisma
```diff
@@ -175,11 +175,38 @@ model EmailAccount {
...
   @@index([lastSummaryEmailAt])
 }
 
+// Stores detected communication patterns for each recipient
+// Used to adapt draft tone based on how the user typically communicates with each person
+model RecipientProfile {
+  id             String   @id @default(cuid())
+  emailAccountId String
+  emailAccount   EmailAccount @relation(fields: [emailAccountId], references: [id], onDelete: Cascade)
+  recipientEmail String
```
elie222, line 190: Do you think it makes sense to store a profile on a recipient by recipient basis, or it would be smarter to store a handful of profile types with instructions for each.

If going with the person by person approach, we could also instruct the ai to analyse the writing style to this specific person while generating the draft.

we're already analysing the email history with the participant in any case. the downside compared to your approach it only has to happen one time and then we're caching it in the database for future use.
the upside is less information needing to be stored in the database. we could end up with a lot of different profiles being created.

thinking about it a little more, it might even be best to store this information in `WritingStyle` prompt.

The reality is that you probably only speak in two or three different ways to different personas. And this could all be stored in a single prompt, and it also means that the user can go and edit it if they want quite easily. If we're storing on a person-by-person basis, it will be really hard for a user to edit it.

My last thought on this is that even with this last approach, it does add some complexity that perhaps we don't need. For me personally, I wouldn't use this feature, and I haven't actually heard a feature request for something like this, but I can understand how it's valuable to certain people. I take it that in your case, you do want ot to write differently for you in different cases?

rsnodgrass: I like the WritingStyle being generalized to just a few different types, I agree individuals probably don't have that many different styles.

For me, I tend to write more formal emails to customers and vendors/financial/legal/etc, and then quick notes to friends or peers on my team. Maybe this isn't as common as I think?

Agreed on avoiding extra database storage where not needed , but perhaps tracking on a per-recipient basic which WritingStyle might mitigate this (Formal, Informal, etc). As you said, Inbox Zero tracks some data per-recipient email already somewhere, for stats/etc. It probably doesn't change for an email once the initial analysis is completed.

I feel like closing this for now is not a bad answer. Wait and see if more people start requesting similar.

rsnodgrass: Btw, based on this thread, Claude recommended:

1. Drop the feature
     Remove RecipientProfile entirely - simplest path, feature isn't triggered anyway
2. Pivot to personas
     Store 2-3 persona types in WritingStyle prompt - user-editable, no extra DB
3. Keep per-recipient
     Fix minor index issue
4. Real-time analysis
     Analyze during draft generation - no storage, but adds latency

elie222: Potentially this could be updated to try and extract multiple styles:
`apps/web/utils/actions/assess.ts`

I do wonder if that makes it messier in cases where people have just one way of writing. But it will make the drafting better for users like yourself, and you're definitely not the only one.

So basically option 2.

Not quite sure what option 3 is.

Option 4 also works, and it's not the latency I'm worried about, but the extra token usage.

rsnodgrass:
I feel like this might be able to be better handled with just a prompt for how to draft replies, rather than a separate tracking feature. 

One could just modify their replies using guidance like:

```
For legal conversations or external emails with @mycompany.com vendors, use a more formal tone.
```

Perhaps that is sufficient (and users can't expand on this as needed).

Most people have separate work/personal emails anyway, so the tone controls are probably less useful if the general reply tone can already be set per email account.

PR #1151, opened 2025-12-31
feat: Pre-Draft All Replies - Automatic draft generation

rsnodgrass:
## Summary

@elie222 This is a feature suggestion, feel free to delete or incorporate.

Automatically generates draft replies for ALL emails detected as needing a reply. Drafts are created directly in the user's Gmail drafts folder - they see them in their normal email client, not in Inbox Zero.

## Changes

- **New model**: `PreDraft` in Prisma schema
  - Tracks generated drafts with status (PENDING, CREATED, SENT, DELETED, FAILED)
  - Stores Gmail draft ID for lifecycle management

- **New settings**: Added to `EmailAccount`
  - `preDraftsEnabled` - Toggle feature on/off (default: false)
  - `preDraftsMaxPerDay` - Daily limit for cost control (default: 10)

- **New file**: `apps/web/utils/pre-drafts/generate-pre-drafts.ts`
  - `generatePreDraftsForAccount()` - Main function to generate drafts
  - Queries ThreadTracker for NEEDS_REPLY threads
  - Respects daily limits and skips already-drafted threads

- **New endpoint**: `POST /api/pre-drafts/generate`
  - Triggers pre-draft generation for an email account
  - Can be called manually or via QStash scheduled job

## How it works

1. User enables pre-drafts in settings (not implemented in this PR)
2. Background job or API call triggers `generatePreDraftsForAccount`
3. System queries ThreadTracker for threads with `type = NEEDS_REPLY`
4. For each thread without a pre-draft:
   - Generates draft content using existing `aiDraftWithKnowledge`
   - Creates draft in Gmail via `draftEmail` API
   - Records draft in `PreDraft` table
5. User opens Gmail, sees drafts ready to send/edit

## Limitations

- Currently Gmail-only (Outlook support can be added later)
- No UI for settings (can be added in follow-up)
- No QStash scheduling configured (needs webhook trigger)

## Test plan

- [ ] Verify PreDraft model works correctly
- [ ] Test with account that has NEEDS_REPLY threads
- [ ] Confirm daily limit is respected
- [ ] Verify drafts appear in Gmail drafts folder

### DIFF
--- a/apps/web/app/api/pre-drafts/generate/route.ts
+++ b/apps/web/app/api/pre-drafts/generate/route.ts
@@ -1,0 +1,24 @@
+import { NextResponse } from "next/server";
+import { withEmailAccount } from "@/utils/middleware";
+import { generatePreDraftsForAccount } from "@/utils/pre-drafts/generate-pre-drafts";
+
+/**
+ * POST /api/pre-drafts/generate
+ * Triggers pre-draft generation for an email account.
+ * Can be called manually or via QStash scheduled job.
+ */
+export const POST = withEmailAccount(async (request) => {
+  const { emailAccountId } = request.auth;
+
+  const results = await generatePreDraftsForAccount(emailAccountId);
+
+  return NextResponse.json({
+    success: true,
+    results,
+    summary: {
+      total: results.length,
+      successful: results.filter((r) => r.success).length,
+      failed: results.filter((r) => !r.success).length,
+    },
+  });
+});
--- a/apps/web/prisma/migrations/20251231000001_add_pre_draft/migration.sql
+++ b/apps/web/prisma/migrations/20251231000001_add_pre_draft/migration.sql
@@ -1,0 +1,30 @@
+-- CreateEnum
+CREATE TYPE "PreDraftStatus" AS ENUM ('PENDING', 'CREATED', 'SENT', 'DELETED', 'FAILED');
+
+-- AlterTable
+ALTER TABLE "EmailAccount" ADD COLUMN "preDraftsEnabled" BOOLEAN NOT NULL DEFAULT false;
+ALTER TABLE "EmailAccount" ADD COLUMN "preDraftsMaxPerDay" INTEGER NOT NULL DEFAULT 10;
+
+-- CreateTable
+CREATE TABLE "PreDraft" (
+    "id" TEXT NOT NULL,
+    "emailAccountId" TEXT NOT NULL,
+    "threadId" TEXT NOT NULL,
+    "messageId" TEXT NOT NULL,
+    "draftId" TEXT,
+    "status" "PreDraftStatus" NOT NULL DEFAULT 'PENDING',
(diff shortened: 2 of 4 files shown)


rsnodgrass:
@elie222 I had some extra Claude Code credits expiring December 31 and wanted to donate them to the project by implementing these features as proof of concepts. Feel free to use, modify, or delete/throw out as you see fit!

**Why Pre-Draft All Replies is a great addition:**

This is the "magic" feature that could be a major time-saver. Instead of users having to click into Inbox Zero to see AI-generated drafts, this creates drafts directly in their Gmail drafts folder. Users open Gmail and find replies already waiting - they just review, edit if needed, and send.

Key benefits:
- **Zero friction**: Drafts appear where users already are (Gmail)
- **Time savings**: The most common action (replying to emails needing response) is pre-done
- **User control**: It's just a draft - users always have final say
- **Cost controlled**: Daily limits (default 10) prevent runaway AI costs

This positions Inbox Zero as truly "zero effort" - the AI does the work proactively, not just on-demand. It's the kind of feature that, once users experience it, they can't go back.

Note: Currently Gmail-only, but the architecture supports adding Outlook later.

elie222:
> This is the "magic" feature that could be a major time-saver. Instead of users having to click into Inbox Zero to see AI-generated drafts, this creates drafts directly in their Gmail drafts folder. Users open Gmail and find replies already waiting - they just review, edit if needed, and send.

Hey, it should already be doing this! Drafting a reply for every needing a reply:

CleanShot 2025-12-31 at 16 54 30

rsnodgrass:
**Update:** After further investigation, I realize this feature is redundant!

Inbox Zero already has this capability built-in:
1. Users set up a rule for "Needs Reply" (TO_REPLY system type)
2. Add a `DRAFT_EMAIL` action to that rule  
3. When emails arrive and are classified as needing a reply, drafts are automatically generated and saved to Gmail

My implementation was essentially duplicating this existing functionality. The infrastructure is already there - users just need to configure their rules appropriately.

I'll close this PR. The other two PRs (Cross-Context Awareness and Recipient-Adaptive Tone) are still valid enhancements to the draft generation pipeline.

elie222:
Got it! We try to set this rule up automatically for users when they go through the onboarding. I'm curious how you missed it? Or did you intentionally turn if off initially?

I ask so we can improve the UX for other users, or if there's a bug somewhere we need to fix.

rsnodgrass:
@elie222 I think there are a few things:

1. When I added a second account, it did not create the REPLY_TO rules at all. And it wasn't clear to me as a user that it was not setup until I just happened to be looking into how the system worked.

2. Given this is such a key feature of Inbox Zero, having it cery clear to users that step is complete. 

Given I think 2 above is an issue, I killed this PR and created a second PR that focused on clearer guidance for users as part of onboarding:  https://github.com/elie222/inbox-zero/pull/1154

PR #1275, opened 2026-01-14
Follow up drafts are too generic

jshwrnr:


apps/web/utils/reply-tracker/generate-draft.ts
```diff
@@ -22,6 +22,9 @@ import {
   getMeetingContext,
   formatMeetingContextForPrompt,
 } from "@/utils/meeting-briefs/recipient-context";
+import { DraftType } from "@/utils/ai/reply/draft-types";
+
+export { DraftType };
```
elie222: Never do import/export. It's just not needed.

The AI puts this is in for no-need. We can add a line to the CLAUDE.md / notes.md that this isn't needed.

apps/web/utils/ai/reply/draft-types.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Type of draft to generate. Each type has different AI prompting behavior.
+ * - DEFAULT: Standard reply draft based on thread context
+ * - FOLLOW_UP: Follow-up reminder when user is waiting for a response
+ */
+export enum DraftType {
```
elie222: This can be placed in the draft-with-knowledge file instead, as it's heavily tied to that.

Also an enum feels like over kill when it could have been:
type DraftType = `default` | `follow-up`

Or even just inlining `default` | `follow-up` as a param. Is enum really necessary?

apps/web/utils/ai/reply/draft-with-knowledge.ts
```diff
@@ -29,6 +28,26 @@ Don't suggest meeting times or mention availability unless specific calendar inf
 Return your response in JSON format.
 `;
 
+const draftTypeInstructions: Record<DraftType, string> = {
```
elie222: I wonder if we should really be overloading here, or we should have 2 different files for this.

Something else that would be helpful to change is LLM "label" that tags what type of call this is.

I wonder if the right approach is to copy and paste this entire file, and have a dedicated function for each. The current approach is limiting because it makes adjusting each harder.

eg. we might want default model for regular drafts, but economy model for follow up drafts.

We'll end up with lots of if/else logic. So I think cleaner is to just have a dedicated file/function for each type of AI call.

Makes it easier to fully adjust the follow up prompt to that exact use case.

elie222:
@jshwrnr fix conflict

apps/web/__tests__/ai/reply/draft-follow-up.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      const result = await aiDraftFollowUp({
+        messages,
+        emailAccount,
+        writingStyle: null,
+      });
+
+      // Check that the result is a non-empty string
+      expect(result).toBeTypeOf("string");
+      if (typeof result === "string") {
+        expect(result.length).toBeGreaterThan(0);
```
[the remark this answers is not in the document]
jshwrnr: Intentionally not adding style assertions - AI outputs are non-deterministic and checking for specific "professional" phrases would make this test flaky. The test verifies the parameter is accepted and produces output; verifying subjective style characteristics in LLM responses isn't reliable for automated testing.

PR #1329, opened 2026-01-19
Display sender names in bulk archive page

jshwrnr:
Display both sender name and email address in bulk archive category cards, matching the behavior in bulk unsubscribe.

**Changes:**
- Fetch sender display names from EmailMessage table for senders without stored names
- Merge names into API response for proper display in UI

The EmailCell component already handles rendering both name and email when available.

### DIFF
--- a/apps/web/utils/actions/categorize.ts
+++ b/apps/web/utils/actions/categorize.ts
@@ -59,6 +59,7 @@ export const bulkCategorizeSendersAction = actionClient
     });
 
     const LIMIT = 100;
+    const MAX_SENDERS = 2000;
 
     let totalUncategorizedSenders = 0;
     let currentOffset: number | undefined = 0;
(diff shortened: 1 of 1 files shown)


apps/web/app/api/user/categorize/senders/categorized/route.ts
```diff
@@ -29,8 +30,37 @@ async function getCategorizedSenders({
...
+        AND "from" IN (${Prisma.join(senderEmails)})
+        AND "fromName" IS NOT NULL
+        AND "fromName" != ''
+      GROUP BY "from"
+    `);
+
+    senderNamesMap = Object.fromEntries(
+      senderNames.map((s) => [s.from, s.fromName || ""]),
+    );
+  }
```
elie222: It shouldn't be using EmailMessage. We may not have stored this data. When we're saving hte Newsletter itself, that's where we want the name stored. No need to map over another database table

elie222: Note, you want to delete all your newsletters in your local db, as it's possible it's implemented but you're seeing old data, and that's why you don't see names.

jshwrnr: Good point - reverted the EmailMessage query. The API now relies on Newsletter.name being populated at creation time. The EmailCell component change (adding text-xs styling) remains to ensure consistent display with bulk unsubscribe.

jshwrnr: After investigation, the categorization flow in `categorizeSender()` doesn't pass `senderName` when calling `updateSenderCategory()`, so `Newsletter.name` is always null. Restored the EmailMessage fallback query to get sender display names.

PR #1350, opened 2026-01-20
Fix account reassignment on organization page

jshwrnr:
## Summary

Fixed a bug where navigating to the organization page would reassign users back to their first email account, even if they had switched to a different account. The issue occurred because the organization page URL doesn't include an emailAccountId parameter.

## How it works

- Track the last known email account ID in React state during the session
- Use this as a fallback when navigating to pages without emailAccountId in the URL
- Fixes AccountSwitcher to properly redirect when switching accounts on pages without account IDs in their path

### DIFF
--- a/apps/web/components/AccountSwitcher.tsx
+++ b/apps/web/components/AccountSwitcher.tsx
@@ -56,10 +56,6 @@ export function AccountSwitcherInternal({
         emailAccountId,
       );
 
-      if (newBasePath === basePath) {
-        return `/${emailAccountId}/automation`;
-      }
-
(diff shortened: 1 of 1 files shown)


apps/web/components/AccountSwitcher.tsx
```diff
@@ -56,6 +56,10 @@ export function AccountSwitcherInternal({
         emailAccountId,
       );
 
+      if (newBasePath === basePath) {
+        return `/${emailAccountId}/automation`;
+      }
```
elie222: what's the idea here?

PR #1393, opened 2026-01-25
Hide announcements for newly signed-up users

jshwrnr:
Set announcementDismissedAt on user signup so newly registered users won't see announcements published before their account was created. Any announcements published after signup will still be shown.

Changes:
- Updated postSignUp hook in auth.ts to set announcementDismissedAt to current timestamp when user signs up
- Existing dismissal logic in announcements API already handles filtering based on this timestamp


apps/web/utils/auth.ts
```diff
@@ -181,6 +181,11 @@ async function postSignUp({
   name?: string | null;
   image?: string | null;
 }) {
+  await prisma.user.update({
```
elie222: i think we can just add this as a default value in the schema if we're going to do it like this. but i think better is just rely on createdAt + announcementDismissedAt in the code itself

jshwrnr: You're saying if there's no `user.announcementDismissedAt`, I should use the `user.createdAt` to determine whether to show the announcement dialog? If so, what should the timeframe be? Like don't show the modal if they sign up less than 24 hours ago? Lmk

elie222: you can use exact same logic. just rely on createdAt

if user signed up on jan 2 it should be treated like announcementDismissedAt = jan 2

```
max(createdAt, announcementDismissedAt)
```

so if announcementDismissedAt is null itll work rely on createdAt

jshwrnr: Got it. Will do that.

PR #1399, opened 2026-01-25
Add team invite step to onboarding and setup pages

jshwrnr:
Auto-create organizations for new users and allow organization owners to invite team members during onboarding.

## Changes
- New onboarding step for team invitations
- Auto-create organizations when users sign up
- Integrate invite modal into setup checklist
- Add organization membership tracking
- Support pending invitations

🤖 Generated with Claude Code

### DIFF
--- a/apps/web/utils/actions/organization.ts
+++ b/apps/web/utils/actions/organization.ts
@@ -127,13 +127,18 @@ export const inviteMemberAction = actionClientUser
         select: { name: true },
       });
 
-      await sendOrganizationInvitation({
-        email,
-        organizationName: org?.name || "Your organization",
-        inviterName:
-          inviterMember.emailAccount.name || inviterMember.emailAccount.email,
-        invitationId: invitation.id,
-      });
+      try {
+        await sendOrganizationInvitation({
+          email,
+          organizationName: org?.name || "Your organization",
+          inviterName:
+            inviterMember.emailAccount.name || inviterMember.emailAccount.email,
+          invitationId: invitation.id,
+        });
+      } catch {
+        await prisma.invitation.delete({ where: { id: invitation.id } });
+        throw new SafeError("Failed to send invitation email");
+      }
     },
   );
 
@@ -544,6 +549,7 @@ export const createOrganizationAndInviteAction = actionClient
           });
           results.push({ email, success: true });
         } catch {
+          await prisma.invitation.delete({ where: { id: invitation.id } });
           results.push({
             email,
             success: false,


apps/web/app/(app)/[emailAccountId]/settings/OrgAnalyticsConsentSection.tsx
```diff
@@ -0,0 +1,64 @@
...
+  );
+
+  const handleToggle = useCallback(
+    (checked: boolean) => {
+      execute({ allowOrgAdminAnalytics: checked });
+    },
+    [execute],
+  );
+
+  if (!isLoading && !data) {
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/settings/OrgAnalyticsConsentSection.tsx]: why early return? this is what <LoadingContent is supposed to do

apps/web/app/(app)/organization/[organizationId]/stats/OrgStats.tsx
```diff
@@ -65,6 +71,33 @@ export function OrgStats({ organizationId }: { organizationId: string }) {
...
+        <ShieldX className="mb-4 h-12 w-12 text-muted-foreground" />
+        <h2 className="mb-2 text-xl font-semibold">Access Denied</h2>
+        <p className="text-muted-foreground">
+          {
+            "You don't have permission to view organization analytics. Only administrators can access this page."
+          }
+        </p>
+      </div>
+    );
+  }
```
Comment (elie222) [inline apps/web/app/(app)/organization/[organizationId]/stats/OrgStats.tsx:92]: building custom components. this should be its own component for reuse. it's not a one off component

Comment (elie222) [reply apps/web/app/(app)/organization/[organizationId]/stats/OrgStats.tsx:92]: also should rely on existing Typography otherwise it leads to inconsistent UIs. see https://getinboxzero.com/components for what's available

apps/web/app/(app)/organization/[organizationId]/stats/page.tsx
```diff
@@ -13,6 +14,10 @@ export default async function OrgStatsPage({
     <PageWrapper>
       <OrganizationTabs organizationId={organizationId} />
 
+      <div className="mt-6">
+        <OrgAnalyticsConsentBanner />
+      </div>
```
Comment (elie222) [inline apps/web/app/(app)/organization/[organizationId]/stats/page.tsx]: now even when banner doesn't show you have the mt-6

apps/web/app/(app)/organization/[organizationId]/OrgAnalyticsConsentBanner.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const isAdmin = hasOrganizationAdminRole(data.role ?? "");
+
+  const title = isAdmin
+    ? "Include your analytics in organization stats"
+    : "Allow organization admins to view your analytics";
+
+  const description = `Your email analytics are currently private. Enable access to let${isAdmin ? " other " : " "}organization admins view your inbox statistics and usage data. This helps your team understand productivity and collaborate more effectively.`;
+
+  return (
+    <ActionCard
```
Comment (elie222) [inline apps/web/app/(app)/organization/[organizationId]/OrgAnalyticsConsentBanner.tsx:53]: you would move that mt-6 here

apps/web/app/(app)/organization/[organizationId]/page.tsx
```diff
@@ -13,6 +14,10 @@ export default async function MembersPage({
     <PageWrapper>
       <OrganizationTabs organizationId={organizationId} />
 
+      <div className="mt-6">
```
Comment (elie222) [inline apps/web/app/(app)/organization/[organizationId]/page.tsx]: mt-6 will always show. it shouldnt be here

apps/web/app/api/organizations/[organizationId]/members/route.ts
```diff
@@ -20,60 +21,65 @@ export const GET = withAuth(
...
-          select: {
-            id: true,
-            name: true,
-            email: true,
-            image: true,
+    },
+    orderBy: [{ role: "asc" }, { createdAt: "asc" }],
+  });
+
+  const pendingInvitations = isAdmin
```
Comment (elie222) [inline apps/web/app/api/organizations/[organizationId]/members/route.ts]: we have promise.all before. why not keep it?

apps/web/app/api/user/organization-membership/route.ts
```diff
@@ -0,0 +1,81 @@
...
+        where: {
+          email: { equals: emailAccount.email, mode: "insensitive" },
+          status: "pending",
+          expiresAt: { gt: new Date() },
+        },
+        select: { id: true },
+      })
+    : null;
+
+  const membership = await prisma.member.findFirst({
```
elie222: could also promise.all here if we want.

apps/web/app/api/user/setup-progress/route.ts
```diff
@@ -27,26 +27,63 @@ async function getSetupProgress({
...
+              _count: {
+                select: {
+                  members: true,
+                  invitations: true,
+                },
+              },
+            },
+          },
+        },
+      },
```
elie222, line 46: just wondering it this makes the query a lot heavier than it was before and if this is really optimal? or perhaps it's not important that it's a little slower

apps/web/components/InviteMemberModal.tsx
```diff
@@ -24,18 +24,83 @@ import {
...
+  const internalState = useDialogState();
+
+  const isControlled = controlledOpen !== undefined;
+  const isOpen = isControlled ? controlledOpen : internalState.isOpen;
+  const onOpenChange = isControlled
+    ? controlledOnOpenChange
+    : internalState.onToggle;
+  const onClose = isControlled
+    ? () => controlledOnOpenChange?.(false)
+    : internalState.onClose;
```
elie222, line 63: isControlled feels like mega hacky code

apps/web/components/InviteMemberModal.tsx
```diff
@@ -24,18 +24,83 @@ import {
...
+          </DialogTrigger>
+        ))}
+
+      <DialogContent className="sm:max-w-md">
+        <DialogHeader>
+          <DialogTitle>Invite Member</DialogTitle>
+          <DialogDescription>
+            {organizationId
+              ? "Send an invitation to join your organization. The recipient will receive an email with instructions."
+              : "Enter email addresses to invite team members. This will create a new organization for your team."}
```
elie222: wondering if this is too verbose

apps/web/components/InviteMemberModal.tsx
```diff
@@ -79,66 +142,134 @@ export function InviteMemberModal({
...
-          <Input
-            type="email"
-            name="email"
-            label="Email Address"
-            placeholder="john.doe@example.com"
-            registerProps={register("email")}
-            error={errors.email}
+      <div className="space-y-2">
+        <div className="flex items-center space-x-2">
+          <label htmlFor="role" className="text-sm font-medium">
```
elie222: we have label as an exported component in other places

apps/web/app/(app)/organization/[organizationId]/stats/page.tsx
Comment (jshwrnr) [reply apps/web/app/(app)/organization/[organizationId]/stats/page.tsx]: good catch

apps/web/app/api/user/setup-progress/route.ts
jshwrnr: Looking at this query, here's my analysis:

Performance Impact:

The query is slightly heavier than just fetching member data, but not significantly:

What's happening:

Fetching only 1 member (take: 1)
Joining to the organization table
Running 2 COUNT aggregations (members + invitations)
Why it's still efficient:

_count in Prisma uses SQL COUNT() - doesn't fetch all records
Only one member is retrieved, not all of them
The counts are aggregations at the database level, which are optimized
Is it optimal?

For a setup progress endpoint, this is perfectly fine because:

It's not a high-frequency endpoint (called once per page load)
The data is small (1 member + 2 integers)
Alternative approaches would be more complex without meaningful performance gains
More optimal alternatives (if you needed them):


// Option 1: Separate query (only if counts were actually slow)
const membership = emailAccount.members[0];
const orgCounts = membership?.organizationId 
  ? await prisma.organization.findUnique({
      where: { id: membership.organizationId },
      select: { _count: { select: { members: true, invitations: true } } }
    })
  : null;

// Option 2: Raw SQL (overkill for this case)
// Only if you had proven performance issues
My recommendation: Keep it as is. The current approach is clean, readable, and the performance impact is negligible for this use case. Only optimize if you measure actual performance problems.

PR #1456, opened 2026-02-03
Add cron scheduled actions

elie222:
Add a cron endpoint to process scheduled actions and keep them pending when QStash is unavailable. Introduce shared handlers and simple fallback routes for digest, summary, and clean workflows to avoid missing endpoints in self-hosted setups.


elie222:
see #1457

PR #1461, opened 2026-02-04
Auto-filing fixes

jshwrnr:
Improves auto-filing folder selection by adding an Add folder action. Keeps folder creation consistent in setup and settings. Refines helper copy for the folder picker.

### DIFF
--- a/apps/web/app/(app)/[emailAccountId]/drive/FilingActivity.tsx
+++ b/apps/web/app/(app)/[emailAccountId]/drive/FilingActivity.tsx
@@ -1,8 +1,10 @@
 "use client";
 
+import { useCallback, useEffect, useState } from "react";
 import { formatDistanceToNow } from "date-fns";
 import { ExternalLinkIcon, InfoIcon } from "lucide-react";
 import { LoadingContent } from "@/components/LoadingContent";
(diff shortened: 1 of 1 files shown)


apps/web/app/(app)/[emailAccountId]/drive/DriveSetup.tsx
```diff
@@ -615,6 +619,7 @@ function SetupFolderSelection({
   const [optimisticFolderIds, setOptimisticFolderIds] = useState<Set<string>>(
     () => new Set(savedFolders.map((f) => f.folderId)),
   );
+  const driveConnectionId = connections[0]?.id ?? null;
```
Comment (elie222) [inline apps/web/app/(app)/[emailAccountId]/drive/DriveSetup.tsx:622]: this will break things if we have multiple drive connections btw

You should at least leave a comment above it. I remember the AI trying to do this in the past as well. 

Like, it's not actually the right approach. It should be working with a specific drive connection ID, and it shouldn't be item zero of the connections array.

Comment (jshwrnr) [reply apps/web/app/(app)/[emailAccountId]/drive/DriveSetup.tsx:622]: Add a TODO for it here: https://github.com/elie222/inbox-zero/pull/1468
If you want to add support for multiple drives, let me know.

PR #1497, opened 2026-02-08
Fix reply-to header preservation in AI assistant replies

jshwrnr:
## Summary
- Add `replyTo` support to `replyToEmail()` in both Gmail and Outlook providers
- Preserve `+ai` and `+assistant` addresses when replying to emails
- Fix filing reply lookup to support multi-turn conversations by walking the thread

## Changes
- Updated `EmailProvider` interface to accept optional `{ replyTo?: string }` parameter
- Modified Gmail and Outlook `replyToEmail()` implementations to set the Reply-To header
- Updated assistant and filing reply handlers to pass the appropriate reply-to address
- Fixed `processFilingReply()` to find filings by walking the entire thread, not just using In-Reply-To header

### DIFF
--- a/apps/web/utils/drive/handle-filing-reply.ts
+++ b/apps/web/utils/drive/handle-filing-reply.ts
@@ -49,7 +49,6 @@ export async function processFilingReply({
     message,
     emailProvider,
     emailAccountId,
-    logger,
   });
 
   if (!filing) {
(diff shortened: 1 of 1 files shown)


apps/web/utils/drive/filing-notifications.ts
```diff
@@ -70,15 +70,21 @@ export async function sendFiledNotification({
       messageHtml,
     });
 
+    const notificationMessageId = await resolveRfc822MessageId({
+      emailProvider,
```
elie222: needed? or can we use messageId like we do in most of the app

PR #1526, opened 2026-02-10
Fix auto-file bug: prevent filing into unapproved folders

jshwrnr:
## Summary

Auto-file was filing documents into folders the user hadn't approved. The AI could return a 'create_new' action for any folder path, and the system would accept it without validation. Additionally, newly created folders were auto-added to the allowed list, silently expanding the user's filing boundaries.

## Changes

- Remove 'create_new' action from AI schema; AI can only use existing allowed folders or skip
- Simplify resolveFolderTarget() to return null for invalid folder IDs instead of creating arbitrary folders  
- Remove folder creation logic from auto-filing path
- Allowed folders list is now a hard boundary, not a soft suggestion

Fixes INB-70

### DIFF
--- a/apps/web/utils/ai/document-filing/analyze-document.ts
+++ b/apps/web/utils/ai/document-filing/analyze-document.ts
@@ -7,21 +7,13 @@ import { cleanExtractedText } from "@/utils/drive/document-extraction";
 const documentAnalysisSchema = z
   .object({
     action: z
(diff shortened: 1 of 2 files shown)


elie222:
This just removes functionality that could be helpful?

The AI should be able to create new groups. eg. place receipts in the folder for the month. It can create March 2026 receipts in this case.

PR #1642, opened 2026-02-19
Fix Anthropic System Message Usage

jayhf:
Anthropic models only allow system prompts to appear at the beginning of a conversation and the current approach fails for Anthropic models after the first message. One message can be sent to the agent, it responds and then an error occurs when attempting to send a second message. There are three ways this can be fixed:

1. Change the additional context to be a user message instead of a system prompt. This is simplest but could potentially affect performance for OpenAI or other models that don't have this restriction
2. Move these messages earlier - this would break the cache optimization
3. Check which model provider is being used and apply (1) conditionally for models with this restriction.

This PR implements (1) for simplicity, but I can go ahead and implement (3) if you prefer


elie222:
Thanks!!

PR #1710, opened 2026-02-25
feat: learn cold email patterns from spam label actions

JasonBates:
## User Story: 

As a self-hosted Inbox Zero user with a Cold Email rule enabled,
I want emails I mark as Junk in my mail client (mail.app) to automatically train the cold email filter,
so that future emails from that sender are blocked at Tier 1 without needing an AI call — and if I accidentally mark something as junk, moving it back to my inbox undoes the learning.

**Why it matters**: Today, cold email detection relies on AI analysis or manual GroupItem configuration. But the strongest signal a user gives is already happening naturally — marking spam in Mail.app/Gmail. This feature closes that feedback loop, making the filter progressively smarter from normal email triage with zero extra effort.

## Summary

- When a user marks an email as junk ( e.g. by marking it as junk in mail.app which adds a SPAM label via Gmail), the sender is saved as a cold email learned pattern with `source=LABEL_ADDED`, so future emails from that sender are caught at Tier 1 without an AI call
- When the user moves an email out of junk (SPAM label removed) and back into the Inbox, the learned pattern is automatically undone — but only patterns created by this handler (`source=LABEL_ADDED`), preserving AI and user-created patterns
- Adds `LABEL_ADDED` to the `GroupItemSource` enum with a corresponding Prisma migration
- NOTE: emails must be moved from junk back to Inbox to trigger the undo 

## Changes

- **New file:** `process-label-added-event.ts` — handles SPAM label additions, finds the Cold Email rule, and saves a learned pattern via `saveLearnedPattern`
- **Modified:** `process-history-item.ts` — routes `LABEL_ADDED` events to the new handler (previously a no-op stub)
- **Modified:** `process-label-removed-event.ts` — adds `undoSpamLearning()` that deletes `LABEL_ADDED`-sourced cold email GroupItems when SPAM is removed; refactored to fetch sender once for both undo and existing label-removal learning
- **Schema:** adds `LABEL_ADDED` to `GroupItemSource` enum

## Design decisions

- Uses `deleteMany` (not `exclude: true`) for undo — this resets to "unknown" so the AI can re-evaluate, rather than permanently marking as "not cold email"
- Scoped to `source: LABEL_ADDED` so undo never affects AI or user-created patterns
- Error handling mirrors existing codebase patterns (inline try/catch with Gmail error type checks)
- No new helper abstractions — follows the existing inline pattern used throughout webhook handlers

## Test plan

- [x] Mark email as junk in Mail.app → verified GroupItem created with `source=LABEL_ADDED`, `exclude=false`
- [x] Move email from Junk to Inbox → verified GroupItem deleted (`deletedCount: 1` in logs, 0 rows in DB)
- [x] Tested with refactored code (single `getMessage` call) deployed via local Docker build
- [ ] Confirm migration applies cleanly on fresh database via `prisma migrate deploy`
- [x] 23 unit tests pass (`process-label-added-event.test.ts`, `process-label-removed-event.test.ts`)

---

## Update: Edge case fix, PII logging, and unit tests

1. **Fix: don't overwrite existing cold email patterns.** If a sender was already classified as a cold email by AI or a user, marking that email as junk would silently overwrite the pattern's source to `LABEL_ADDED`. Then if the user later un-junked it, the undo logic would delete the pattern entirely — destroying the original AI classification. Now we check if the sender already exists in the cold email group before saving, and skip if so.

2. **Fix: PII logging level.** Sender email addresses were logged at `info` level in both label event handlers. Moved to `trace` to follow the project's logging conventions for PII fields.

3. **Add unit tests.** After reviewing the [contributing guide](https://docs.getinboxzero.com/contributing), added 23 tests across two files covering the full spam learning flow — label added, label removed, undo logic, edge cases (missing IDs, rate limits, deleted messages), and the new guard check.

### DIFF
--- a/apps/web/app/api/google/webhook/process-label-added-event.ts
+++ b/apps/web/app/api/google/webhook/process-label-added-event.ts
@@ -1,5 +1,9 @@
 import type { gmail_v1 } from "@googleapis/gmail";
-import { GroupItemSource, GroupItemType } from "@/generated/prisma/enums";
+import {
+  GroupItemSource,
+  GroupItemType,
+  SystemType,
+} from "@/generated/prisma/enums";
 import { saveLearnedPattern } from "@/utils/rule/learned-patterns";
 import { extractEmailAddress } from "@/utils/email";
 import type { EmailAccountWithAI } from "@/utils/llms/types";
@@ -57,7 +61,7 @@ export async function handleLabelAddedEvent(
   const coldEmailRule = await prisma.rule.findFirst({
     where: {
       emailAccountId,
-      systemType: "COLD_EMAIL",
+      systemType: SystemType.COLD_EMAIL,
       enabled: true,
     },
     select: { id: true, groupId: true },
--- a/apps/web/app/api/google/webhook/process-label-removed-event.ts
+++ b/apps/web/app/api/google/webhook/process-label-removed-event.ts
@@ -3,6 +3,7 @@ import {
   ActionType,
   GroupItemSource,
   GroupItemType,
+  SystemType,
 } from "@/generated/prisma/enums";
 import { extractEmailAddress } from "@/utils/email";
 import type { EmailAccountWithAI } from "@/utils/llms/types";
@@ -212,7 +213,7 @@ async function undoSpamLearning({
   const coldEmailRule = await prisma.rule.findFirst({
     where: {
       emailAccountId,
-      systemType: "COLD_EMAIL",
+      systemType: SystemType.COLD_EMAIL,
       enabled: true,
     },
     select: { id: true, groupId: true },


elie222:
thanks! taking a look!

JasonBates:
**Note: Gmail / Mail.app only**

This works because Gmail sends explicit LABEL_ADDED/LABEL_REMOVED events — we know exactly what changed and when.

Outlook would need separate work because:
1. Outlook's "mark as junk" is a folder move (parentFolderId change), not a label — the webhook subscription (`/me/messages`) fires, but the handler's scope check (`process-history.ts:76`) drops messages not in Inbox or Sent
2. Outlook webhooks only say "message updated" with no detail on what changed — you'd have to infer "just marked as junk" from current state, with no "before" to compare against
3. The existing `learnFromOutlookLabelRemoval` only queries `ActionType.LABEL` actions, but most system categories use `MOVE_FOLDER` on Outlook (see #1744)

elie222:
Sorry for delay on this one! Hope to get it merged in today!

elie222:
Thanks!

PR #1817, opened 2026-03-06
fix(outlook): learn when MOVE_FOLDER classification is undone

### LINKED ISSUE #1744: POTENTIAL BUG: Outlook label-removal learning doesn't trigger for MOVE_FOLDER categories
## What might be broken

On Outlook, when Inbox Zero classifies an email (e.g. as Cold Email) and moves it to a folder, and the user then drags it back out to correct the classification — the system currently might not learn from that correction. The learning code only looks for category (label) changes, but on Outlook most system rules use folder moves instead.

I don't have Outlook so I can't test this. Found it while reading through the code for a different feature and asking Claude Code a lot of questions.

## Why

Claude code tells me that Outlook has two separate things: **categories** (metadata tags on a message, like Gmail labels) and **folders** (where the message physically lives). They're unrelated in the Microsoft Graph API — see [message resource docs](https://learn.microsoft.com/en-us/graph/api/resources/message?view=graph-rest-1.0).

In `consts.ts`, most system rules set `categoryActionMicrosoft: "move_folder"` — meaning Cold Email, Newsletter, Marketing, Receipt, and Notification all move emails into folders on Outlook.

But `learnFromOutlookLabelRemoval` only queries for `ActionType.LABEL` actions (line 35 of `learn-label-removal.ts`) and checks `message.labelIds` which maps to categories. It never looks at `ActionType.MOVE_FOLDER` or checks `parentFolderId`.

So the rule-matching query might come back empty and learning silently wouldn't happen.

Apparently Calendar is the exception — it uses categories on Outlook, so learning should work for that one.

## To fix .... theoretically ¯\_(ツ)_/¯ 

1. Also query for `ActionType.MOVE_FOLDER` actions in the executed rules lookup
2. For those actions, check whether the message is still in the target folder (`parentFolderId`) rather than checking category presence

( strangest bug report I ever made. )

SergioChan:
## Summary
- include `MOVE_FOLDER` actions when loading executed Outlook system-rule actions for removal learning
- treat a moved message as "removed" when the current `parentFolderId` no longer matches the rule action's target folder
- persist `parentFolderId` on parsed Outlook messages and add coverage for both mismatch/match folder cases

## Testing
- Attempted: `pnpm --dir inbox-zero --filter inbox-zero-ai test -- --run apps/web/app/api/outlook/webhook/learn-label-removal.test.ts`
- Not runnable in this environment: repo dependencies are not installed (`node_modules missing`) and the script fails before test execution with `cross-env: Permission denied`.

## Related
Fixes #1744

### DIFF
--- a/apps/web/app/api/outlook/webhook/learn-label-removal.test.ts
+++ b/apps/web/app/api/outlook/webhook/learn-label-removal.test.ts
@@ -175,6 +175,33 @@ describe("learnFromOutlookLabelRemoval", () => {
     expect(saveLearnedPattern).not.toHaveBeenCalled();
   });
 
+  it("does not treat missing label snapshot as label removal", async () => {
+    vi.mocked(prisma.executedRule.findMany).mockResolvedValue([
+      {
+        rule: {
+          id: "rule-1",
+          systemType: SystemType.NEWSLETTER,
+        },
+        actionItems: [{ labelId: "label-newsletter", label: "Newsletter" }],
+      },
+    ] as any);
+
+    const message = getMockParsedMessage({
+      id: "message-123",
+      threadId: "thread-123",
+      labelIds: undefined,
+      headers: { from: "sender@example.com" },
+    });
+
+    await learnFromOutlookLabelRemoval({
+      message,
+      emailAccountId: "email-account-123",
+      logger,
+    });
+
+    expect(saveLearnedPattern).not.toHaveBeenCalled();
+  });
+
   it("does not learn for non-learnable rule types", async () => {
     vi.mocked(prisma.executedRule.findMany).mockResolvedValue([
       {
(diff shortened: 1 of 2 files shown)


SergioChan:
Follow-up on the review note identified by cubic: fixed in 3fc8f82.

What changed:
- Guarded label-removal inference behind a real labels snapshot (Array.isArray(message.labelIds)), so missing webhook labels are treated as unknown state, not empty labels.
- Added regression test: "does not treat missing label snapshot as label removal".

Validation:
- Attempted: pnpm --dir /home/chen/.openclaw/workspace/inbox-zero exec vitest apps/web/app/api/outlook/webhook/learn-label-removal.test.ts
- Blocked locally by missing tsconfig preset dependency resolution in this environment (failed to resolve extends: tsconfig/nextjs.json).

Happy to rerun once workspace deps/toolchain are fully available.

elie222:
Thanks!

PR #1972, opened 2026-03-18
assistant-chat: add mobile confirm-email-action route

jshwrnr:
Adds a dedicated authenticated API route for confirming assistant email actions from mobile clients. It mirrors server action behavior and validates payloads with the shared schema.

- add POST /api/chat/confirm-email-action using withEmailAccount
- call confirmAssistantEmailActionForAccount with account/provider context
- document that mobile-native integrations may use POST API routes in AGENTS.md

### DIFF
--- a/apps/web/app/api/chat/confirm-email-action/route.ts
+++ b/apps/web/app/api/chat/confirm-email-action/route.ts
@@ -17,7 +17,13 @@ export const POST = withEmailAccount(
       return NextResponse.json({ error: "Not authenticated" }, { status: 401 });
     }
 
-    const json = await request.json();
+    let json: unknown;
+    try {
+      json = await request.json();
+    } catch {
+      return NextResponse.json({ error: "Invalid JSON body" }, { status: 400 });
+    }
+
     const { data, error } = confirmAssistantEmailActionBody.safeParse(json);
     if (error) {
       return NextResponse.json({ error: error.errors }, { status: 400 });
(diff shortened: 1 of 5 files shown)


apps/web/utils/redis/account-validation.ts
```diff
@@ -63,6 +72,14 @@ export async function invalidateAccountValidation({
   userId: string;
   emailAccountId: string;
 }): Promise<void> {
+  if (!isRedisConfigured()) return;
```
elie222: The AI probably put this in, but remove all this Redis stuff. It's not relevant for this PR

elie222: Removed the unrelated Redis and local-bypass changes from this branch so the PR stays scoped to the confirm-email route work.

apps/web/app/api/chat/confirm-email-action/route.ts
```diff
@@ -0,0 +1,39 @@
...
+  "chat/confirm-email-action",
+  async (request) => {
+    const emailAccountId = request.auth.emailAccountId;
+
+    const user = await getEmailAccountWithAi({ emailAccountId });
+    if (!user) {
+      return NextResponse.json({ error: "Not authenticated" }, { status: 401 });
+    }
+
+    const json = await request.json();
```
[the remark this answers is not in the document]
elie222: Addressed. Cubic identified a valid malformed-body path here, so the route now catches `request.json()` parse failures and returns `400` before schema validation. I also added a colocated route test that covers malformed JSON and the normal confirm path.

PR #1998, opened 2026-03-23
chats: add update and soft-delete APIs

jshwrnr:
Adds chat metadata fields and API support for updating and soft-deleting chats.

- Add Chat.title, Chat.starred, and Chat.deletedAt with migration and index.
- Add PATCH /api/chats/[chatId] with zod validation for title and starred.
- Add DELETE /api/chats/[chatId] as a soft delete.
- Exclude soft-deleted chats from list and detail reads.

### DIFF
--- a/apps/web/app/api/chats/[chatId]/route.test.ts
+++ b/apps/web/app/api/chats/[chatId]/route.test.ts
@@ -31,11 +31,11 @@ describe("PATCH /api/chats/[chatId]", () => {
     vi.clearAllMocks();
   });
 
-  it("rejects starred-only payloads", async () => {
+  it("returns 400 when only unknown fields are provided", async () => {
     const request = new Request("http://localhost/api/chats/chat-1", {
       method: "PATCH",
       headers: { "content-type": "application/json" },
-      body: JSON.stringify({ starred: true }),
+      body: JSON.stringify({ unsupportedField: true }),
     });
 
     const response = await PATCH(request as never, {


apps/web/prisma/schema.prisma
```diff
@@ -976,6 +976,10 @@ model Chat {
   createdAt DateTime @default(now())
   updatedAt DateTime @updatedAt
 
+  title     String?
+  starred   Boolean   @default(false)
```
elie222: why did we add starred?

jshwrnr: Added so that mobile chat threads can be starred/deleted.

apps/web/prisma/migrations/20260323170000_chat_title_star_delete/migration.sql
```diff
@@ -0,0 +1,9 @@
+-- AlterTable
+ALTER TABLE "Chat"
+ADD COLUMN "title" TEXT,
```
elie222: should we call it name?

apps/web/prisma/migrations/20260323170000_chat_title_star_delete/migration.sql
```diff
@@ -0,0 +1,9 @@
...
+ADD COLUMN "starred" BOOLEAN NOT NULL DEFAULT false,
```
elie222: we can just do this locally in the app. may add to db in the future

PR #2244, opened 2026-04-14
fix: respect high confidence threshold on draft generation (INB-176)

gentlemandev:
## Summary

- With Draft confidence set to "High confidence", the assistant was still drafting replies for emails that required the user's personal judgment (opinions, commitments, private details). The draft gate already supports skipping low-confidence drafts, but the model was rating fluent-but-speculative drafts as `HIGH_CONFIDENCE` because the prompt gave it almost no guidance on how to self-rate.
- Add an explicit confidence self-assessment section to the draft system prompt. It calibrates the three levels and enumerates, at a concept level, the cases that must be rated as low confidence (subjective judgment, personal decisions, private info, facts not grounded in context, answers that would require fabrication).
- Tighten the Zod schema description to mirror the prompt guidance.
- Bump `DRAFT_PIPELINE_VERSION` from 7 -> 8 so downstream analytics attribute reply quality under the new pipeline.

No keyword/regex rules were added; the fix stays AI-first and relies on the existing `meetsDraftReplyConfidenceRequirement` gate in `utils/reply-tracker/generate-draft.ts`.

## Test plan

- [x] `pnpm test utils/ai/reply/draft-confidence.test.ts utils/ai/reply/draft-reply.formatting.test.ts utils/reply-tracker/generate-draft.test.ts utils/actions/generate-reply.test.ts utils/redis/reply.test.ts` passes (43 tests).
- [ ] Manual QA: with Draft confidence = High confidence, verify that emails requiring personal judgment (subjective opinions, private details) no longer produce drafts, while clearly grounded replies still do.
- [ ] Consider adding an eval suite for confidence self-rating on subjective vs grounded prompts as a follow-up.

Linear: INB-176

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/apps/web/utils/ai/reply/draft-attribution.ts
+++ b/apps/web/utils/ai/reply/draft-attribution.ts
@@ -1,7 +1,7 @@
 // Bump this when draft output behavior changes in a way that would affect
 // quality comparisons, including prompt, retrieval, routing, or
 // post-processing changes.
-export const DRAFT_PIPELINE_VERSION = 7;
+export const DRAFT_PIPELINE_VERSION = 8;
 
 export type DraftAttribution = {
   provider: string;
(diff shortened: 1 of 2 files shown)


gentlemandev:

PR #2245, opened 2026-04-14
fix: re-run AI attachment selection on draft cache miss

gentlemandev:
## Summary

Fixes INB-146. When a `DRAFT_EMAIL` action runs for a rule with AI-selected attachment sources configured (Drive folder / file) and the Redis-cached draft is missing, the action silently dropped all attachments even if the source was fully indexed and the email explicitly requested a matching file.

The two-step design stores the AI-selected attachments alongside the draft during generation (`selectDraftAttachmentsForRule` -> Redis), then re-reads them when the action executes. On cache miss (scheduled/delayed actions, expired cache, rules configured after draft generation, etc.) `getDraftSelectedAttachments` returned `[]`, so users with unindexed/recently-changed folders - or any timing-skew between draft generation and action execution - silently got no attachments.

## Change

- On cache miss in `apps/web/utils/ai/action-attachments.ts`, fall back to running `selectDraftAttachmentsForRule` inline using the current email content.
- The selection function already short-circuits when the rule has no attachment sources configured, so the extra LLM call only happens for rules that actually expect attachments.
- Updated the existing "cache miss" unit test to assert the new fallback behavior (attachments are still selected and resolved) and added a guard on the cache-hit path to ensure we don't double-run selection.

## Test plan

- [x] `pnpm test utils/ai/actions.test.ts` - 11/11 passing
- [x] `pnpm test utils/attachments utils/reply-tracker/generate-draft.test.ts` - 17/17 passing
- [ ] Manual: configure a rule with a Drive folder source, trigger a scheduled/delayed DRAFT_EMAIL action after the cache has expired, confirm the matching PDF is attached.

Refs INB-146.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/apps/web/utils/ai/action-attachments.ts
+++ b/apps/web/utils/ai/action-attachments.ts
@@ -4,10 +4,15 @@ import {
   type SelectedAttachment,
   attachmentSourceInputSchema,
 } from "@/utils/attachments/source-schema";
-import { resolveDraftAttachments } from "@/utils/attachments/draft-attachments";
+import {
+  resolveDraftAttachments,
+  selectDraftAttachmentsForRule,
+} from "@/utils/attachments/draft-attachments";
 import type { ActionItem, EmailForAction } from "@/utils/ai/types";
 import type { Logger } from "@/utils/logger";
 import { getReplyWithConfidence } from "@/utils/redis/reply";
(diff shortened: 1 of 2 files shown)


gentlemandev:

PR #2246, opened 2026-04-14
assistant: fix chat label_threads removing existing labels (INB-137)

gentlemandev:
## Summary

Fixes INB-137: asking the assistant to "label all emails from X as Y" previously resulted in emails losing their existing label and being marked read, with no new label applied.

Root cause: when the requested label didn't exist, `manageInbox`'s `label_threads` action threw an error pointing at `createOrGetLabel`. The model would then fall back to `archive_threads` / `mark_read_threads` to "clean up" instead of chaining `createOrGetLabel` + `label_threads`. System prompt also omitted labeling from core responsibilities, which made this fallback more likely.

- `label_threads` now auto-creates a missing label (same pattern used by `archive_threads` via `resolveLabelNameAndId`), so a single tool call suffices.
- Clarified the `manageInbox` tool and action descriptions so the model treats `label_threads` as a non-destructive add-only action and uses it for "label X as Y" requests.
- Added labeling to the assistant's listed core responsibilities.

No change to the underlying email-provider `labelMessage` implementations — both Gmail and Outlook were already additive.

## Test plan

- [x] `pnpm test chat-inbox-tools` (10 passed) — updated the "unknown labelName" test to assert auto-create behavior.
- [x] `pnpm test chat-label-tools` (4 passed) — no behavioral change to `createOrGetLabel` tool.
- [ ] Manual QA: ask chat "label all emails from X as 'NewLabel'" and verify (1) the new label is created, (2) it is applied to matching threads, (3) existing labels remain, (4) threads are not archived or marked read.

### DIFF
--- a/apps/web/utils/ai/assistant/chat-inbox-tools.test.ts
+++ b/apps/web/utils/ai/assistant/chat-inbox-tools.test.ts
@@ -413,14 +413,22 @@ describe("chat inbox tools", () => {
     expect(maxInFlight).toBeLessThanOrEqual(3);
   });
 
-  it("returns a descriptive error when label_threads receives an unknown labelName", async () => {
-    const getThreadMessages = vi.fn();
+  it("auto-creates a missing label when label_threads receives an unknown labelName", async () => {
+    const getThreadMessages = vi
+      .fn()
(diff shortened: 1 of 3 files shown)


gentlemandev:

PR #2247, opened 2026-04-14
fix: allow closing chat panel and auto-reset on navigation (INB-135)

gentlemandev:
## Summary
- Adds a visible close (X) button to the AI Chat sidebar header
- Auto-closes the chat panel when the route changes so it no longer persists across unrelated pages
- Ensures the close action clears the sidebar state (used by the shared layout to shift content)

## Root cause
`SidebarRight` rendered `Chat` inside a right sidebar whose open state lives in the shared `SidebarProvider` context (persisted via cookie). There was no close affordance in the chat header, and nothing reset the sidebar state on navigation, so once opened from `/automation` it stayed open on every other app route.

## Fix
- `components/SidebarRight.tsx`: wire a `handleClose` callback that removes the sidebar from the provider's open state (or closes the mobile sheet), close on pathname change, and close on unmount so remounting on another route starts fresh.
- `components/assistant-chat/chat.tsx`: accept an optional `onClose` prop and render an X button in `ChatTopBar` when provided. The full-page assistant (`/assistant`) does not pass `onClose`, so the X only appears when the chat is in the dismissible side panel.

## Test plan
- [ ] Open `/automation`, click `AI Chat`: panel opens, content shifts, X button visible in header
- [ ] Click X: panel closes and content reflows
- [ ] Open panel, navigate to another app route: panel auto-closes
- [ ] Navigate to `/assistant`: full-page chat renders without an X button (expected)
- [ ] On mobile, open via toggle then close via X: mobile sheet closes

### DIFF
--- a/apps/web/components/SidebarRight.tsx
+++ b/apps/web/components/SidebarRight.tsx
@@ -1,5 +1,7 @@
 "use client";
 
+import { useCallback, useEffect, useRef } from "react";
+import { usePathname } from "next/navigation";
 import { useSidebar } from "@/components/ui/sidebar";
 import { Chat } from "@/components/assistant-chat/chat";
 import { cn } from "@/utils";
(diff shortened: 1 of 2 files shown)


gentlemandev:

PR #2248, opened 2026-04-14
feat: warn when rule uses Drive attachments but Drive is disconnected (INB-143)

gentlemandev:
## Summary

Fixes INB-143. When a rule is configured with Drive-backed attachments (AttachmentSource rows or staticAttachments on Actions) and the Drive connection is later disconnected, the rule keeps firing and replies are sent with no attachments, silently. This adds a warning surface on the rules list.

- `GET /api/user/rules` now computes `hasDisconnectedDriveAttachments` per rule. A rule is flagged if any `AttachmentSource.driveConnection.isConnected` is false, or if any `Action.staticAttachments[].driveConnectionId` does not match a currently connected `DriveConnection`.
- The Rules table renders a warning icon next to the rule name that links to the Drive settings page, with a tooltip explaining the attachments will be skipped until the Drive connection is restored.

## Test plan

- [x] Existing `Rules.test.tsx` cases still pass.
- [x] New test verifies the warning badge appears when `hasDisconnectedDriveAttachments` is true.
- [ ] Manually disconnect a Drive and confirm the warning appears on any rule that references it (both AttachmentSource and staticAttachments paths).
- [ ] Follow-up (not in this PR): inline alert inside the rule editor; execution-history entry noting the skipped attachments.

### DIFF
--- a/apps/web/app/(app)/[emailAccountId]/assistant/Rules.test.tsx
+++ b/apps/web/app/(app)/[emailAccountId]/assistant/Rules.test.tsx
@@ -25,10 +25,15 @@ vi.mock("next/link", () => ({
   default: ({
     children,
     href,
+    ...rest
   }: {
     children: React.ReactNode;
     href: string;
-  }) => <a href={href}>{children}</a>,
+  } & React.AnchorHTMLAttributes<HTMLAnchorElement>) => (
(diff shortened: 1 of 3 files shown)


gentlemandev:

PR #2254, opened 2026-04-14
fix: re-run AI attachment selection on draft cache miss (INB-146)

gentlemandev:
## Summary
- Fixes INB-146: rules configured with AI-selected Drive sources generated drafts with no attachments whenever the Redis draft cache was missing (cache expiry, replay after restart, etc.).
- `getDraftSelectedAttachments` now falls back to `selectDraftAttachmentsForRule` on cache miss instead of silently returning `[]`.
- The fallback is cost-conscious: `selectDraftAttachmentsForRule` already short-circuits with a cheap DB check when the rule has no attachment sources, so there is no extra LLM call for rules without Drive sources. Only rules that genuinely have AI-selected sources incur the selection call.

## Test plan
- [x] `pnpm --filter inbox-zero-ai test utils/ai/actions.test.ts` — inverted existing wrong test ("skips draft attachments when the rule cache is missing") into "re-runs AI attachment selection when the rule cache is missing"; now passes.
- [x] `pnpm --filter inbox-zero-ai test utils/reply-tracker/generate-draft.test.ts utils/attachments/` — 17 tests pass (no regressions).
- [ ] Manual: verify that a reply-tracker rule with a Drive folder source attaches the relevant PDF on a fresh draft when the Redis cache has expired.

Follow-ups
- Consider passing the already-loaded `emailAccount` through the action pipeline to avoid the extra DB read on cache miss.

### DIFF
--- a/apps/web/utils/ai/action-attachments.ts
+++ b/apps/web/utils/ai/action-attachments.ts
@@ -4,10 +4,17 @@ import {
   type SelectedAttachment,
   attachmentSourceInputSchema,
 } from "@/utils/attachments/source-schema";
-import { resolveDraftAttachments } from "@/utils/attachments/draft-attachments";
+import {
+  resolveDraftAttachments,
+  selectDraftAttachmentsForRule,
+} from "@/utils/attachments/draft-attachments";
 import type { ActionItem, EmailForAction } from "@/utils/ai/types";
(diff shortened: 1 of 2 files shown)


elie222:
Merged in https://github.com/elie222/inbox-zero/pull/2270

PR #2314, opened 2026-04-20
Add Delete and Spam buttons to Slack notification cards

alex-brecher:
## Summary

Adds two buttons — **Delete** and **Spam** — to the notify-mode Slack card, alongside the existing Archive and Mark read buttons. Draft-mode cards (Send reply / Edit draft / Dismiss) are unaffected.

## Why

When the `Slack` action notifies a user about a matched email, the current card supports only Archive and Mark read. For rules that catch spammy or clearly deletable senders, users want to trash or mark-as-spam straight from Slack without jumping to Gmail/Outlook.

## What changed

Single-file change to `apps/web/utils/messaging/rule-notifications.ts`:

- Two new action IDs: `rule_notify_trash` and `rule_notify_mark_spam`
- Registered in `RULE_NOTIFICATION_ACTION_IDS` so the Slack adapter routes clicks back to the handler
- Switch cases wired to `handleTrashNotification` and `handleMarkSpamNotification`
- Handlers call `provider.trashThread(threadId, email, "user")` and `provider.markSpam(threadId)` respectively, then edit the Slack card to the terminal "Moved to trash." / "Marked as spam." state (same pattern as the existing Archive/Mark-read handlers)
- Two new `Button(...)` entries appended to the `Actions` block in notify mode

Works for both Gmail and Microsoft providers — the `EmailProvider` interface already exposes `trashThread` and `markSpam` for both backends.

## Test plan

- [x] Triggered rule on Gmail account, clicked Delete on Slack card → thread moved to Gmail Trash, card updated to "Moved to trash."
- [x] Triggered rule on Gmail account, clicked Spam → thread moved to Spam label, card updated to "Marked as spam."
- [x] Existing Archive and Mark read buttons still work
- [x] Draft-mode cards (Send reply / Edit draft / Dismiss) unchanged

## Notes

Kept the handler pattern identical to `handleMarkReadNotification` for review-friendliness. No schema changes, no config, no env vars.

### DIFF
--- a/apps/web/utils/messaging/rule-notifications.test.ts
+++ b/apps/web/utils/messaging/rule-notifications.test.ts
@@ -149,7 +149,7 @@ describe("handleRuleNotificationAction", () => {
       raw: { team: { id: "team-1" } },
       threadId: "slack-thread-1",
       messageId: "slack-message-1",
-      adapter: { editMessage },
+      adapter: { name: "slack", editMessage },
       thread: { postEphemeral: vi.fn() },
     } as any;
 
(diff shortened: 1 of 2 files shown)


elie222:
Thanks!

PR #2401, opened 2026-04-27
history: Group entries by thread

elie222:
Groups automation history and test results by thread so related messages stay together. Adds expandable rows that keep the default view compact while exposing per-message handling details.

- Changes history pagination to use thread groups
- Groups Test tab messages by thread with expandable per-message rows
- Keeps per-message rule actions, fix controls, and rerun/test controls available


elie222:
Replaced by #3660, which implements History conversation grouping on current main with lazy, paginated message details, database-side group counting, preserved date headings and Fix controls, and browser/API regression coverage. Test-tab expansion is deferred as a separate change.

PR #2445, opened 2026-04-29
Fix mailbox count handling

elie222:
Adds a provider-backed mailbox count tool so assistant responses can use exact folder, label, and category totals instead of inferring from search pages.

- Uses provider metadata and count APIs for mailbox totals.
- Preserves folder count metadata and adds focused unit coverage.


elie222:
Closing because this implementation approach is not being pursued.

PR #2580, opened 2026-05-09
CLASS-09: PROMOTIONS shortcut as post-loop fallback (deferred)

rebekah-create:
## Status: deferred

Decision (2026-05-09): revert CLASS-09 from `main` and park this redesign on a draft PR. Re-evaluate later as a potential future feature.

## Background

CLASS-09 (commit ce3492916) added a Gmail PROMOTIONS shortcut intended to skip the LLM for obviously promotional mail. The implementation lived inside the per-rule loop and pushed Marketing into `matches[]` alongside any other rule that matched. With `multiRuleSelectionEnabled=false`, two PROMOTIONS-flagged emails were observed to be classified twice on 2026-05-09:

- TD Furn digest 9:00am ET → Marketing + TD Furn (learned-pattern hit)
- EVgo 9:35am ET → Marketing + Uncertain (LLM still ran because no learned pattern hit)

## What this branch changes

- Removes the in-loop PROMOTIONS check.
- Adds a true fallback after the rule loop and `hasLearnedPatternMatch` calculation: if `matches.length === 0 && !hasLearnedPatternMatch && hasOnlyPromotionsCategory(message.labelIds)` and a Marketing system rule exists, return Marketing solo and skip the LLM. Otherwise the existing match path runs unchanged.
- Strengthens the existing learned-pattern test to assert Marketing is absent from `matches[]`. Adds a parallel test for the static-match case.

## Behavior change worth naming

PROMOTIONS-only emails with no other static or learned-pattern match now bypass the LLM entirely. That means conversation-tracker rules (TO_REPLY, AWAITING_REPLY, FYI, ACTIONED) and other AI rules like Uncertain are not evaluated for these messages.

This matches the original cost-saving intent but represents a real divergence from upstream behavior. The argument for it: PROMOTIONS-only senders are overwhelmingly noreply/newsletter senders, which `filterConversationStatusRulesWithMetadata` already strips. The argument against: any AI-only rule the user might add later will be silently bypassed for promotional mail.

## Why deferred

The if/and/and predicate gates important downstream behavior. Even with tests covering the obvious cases, there's residual risk that a future rule shape (e.g. a non-Marketing AI rule the user adds with the intent of catching promotional mail) would be silently bypassed. Worth more thinking before shipping.

## To pick this back up

1. Re-run the tests in `apps/web/utils/ai/choose-rule/match-rules.test.ts`.
2. Confirm the behavior change is still desired (or revisit whether the LLM should still run on PROMOTIONS-only mail when AI rules exist).
3. Rebase on current `main` (CLASS-09 was reverted).
4. Mark ready for review.


rebekah-create:
Sorry — opened against the wrong repo by mistake. Closing.

PR #2620, opened 2026-05-11
Disable digest schedule controls in settings page

rebekah-create:
## Summary
- Follow-up to #22, which greyed out the wrong form.
- PR #22 disabled `DigestScheduleForm.tsx`, which is only rendered in the onboarding flow (`StepDigest.tsx`, `StepDigestV1.tsx`).
- The live settings page at `/assistant/settings` → Digest renders `DigestSettingsForm.tsx` — a separate, duplicated implementation whose frequency `Select`, day-of-week `Select`, and `TimePicker` were never touched. So production still let the user edit the digest time despite the server-side 09:00 ET cron override.

## Fix
- Mirror PR #22 on `DigestSettingsForm.tsx`: `opacity-60 pointer-events-none` wrapper, `disabled` on each control, info tooltip on the "Send the digest email" label with the same `SERVER_SIDE_OVERRIDE_TOOLTIP` copy.
- Submission is unchanged. The Save button still saves digest items; the schedule values it re-sends are identical to the fetched values (user can't edit them), and the cron container ignores `Schedule` anyway.

## Follow-up worth considering (not in this PR)
The two forms duplicate ~150 lines of identical schedule UI. Worth extracting a shared `<ScheduleFields disabled />` component the next time someone touches this area — would have prevented the original miss.

## Test plan
- [ ] Load `/[emailAccountId]/assistant/settings` → Digest tab; confirm frequency/day/time controls are greyed out, not clickable, and the label shows an info tooltip on hover.
- [ ] Confirm onboarding (`StepDigest`) still looks the same as before.
- [ ] Click Save with no changes; expect success toast and no schedule mutation visible in `Schedule.updatedAt` (or harmless no-op).


rebekah-create:
Opened against the wrong remote — this is a self-hosted-fork fix, not relevant upstream. Sorry for the noise.

PR #2634, opened 2026-05-12
feat(login): add NEXT_PUBLIC_LOGIN_PROVIDERS env var to control visible login options

Ashut0sh-mishra:
## Summary

Adds `NEXT_PUBLIC_LOGIN_PROVIDERS`, an optional comma-separated allowlist of login providers to render on the login screen. Valid tokens (case-insensitive): `google`, `microsoft`, `apple`, `sso`.

This solves the access-control problem described in #2106: self-hosted SSO-only deployments currently can't hide the Google / Microsoft buttons, so anyone with a Google or Microsoft account can self-provision and bypass organisational access control.

## Behaviour

- **Unset / empty** &rarr; show all providers (no behaviour change for existing deployments).
- **`NEXT_PUBLIC_LOGIN_PROVIDERS=sso`** &rarr; SSO-only lockdown.
- **`NEXT_PUBLIC_LOGIN_PROVIDERS=google,microsoft`** &rarr; hide SSO button.
- **`NEXT_PUBLIC_LOGIN_PROVIDERS=microsoft,sso`** &rarr; hide Google button.
- The env var can **only narrow** what is otherwise enabled — it is intersected with `hasMicrosoftOauthConfig()`, `env.SSO_LOGIN_ENABLED`, and `env.NEXT_PUBLIC_SHOW_APPLE_LOGIN`, so it can never widen the visible buttons.
- If every token in the list is unknown, the helper falls back to "show all" to avoid locking admins out of their own deployment.

## Changes

- [`apps/web/env.ts`](https://github.com/elie222/inbox-zero/blob/main/apps/web/env.ts): register `NEXT_PUBLIC_LOGIN_PROVIDERS` (optional string).
- [`apps/web/app/(landing)/login/login-providers.ts`](https://github.com/elie222/inbox-zero/blob/main/apps/web/app/(landing)/login): pure parser + helper, exported for reuse.
- [`apps/web/app/(landing)/login/page.tsx`](https://github.com/elie222/inbox-zero/blob/main/apps/web/app/(landing)/login/page.tsx): wires `enabledProviders` into `<LoginForm>`.
- [`apps/web/app/(landing)/login/LoginForm.tsx`](https://github.com/elie222/inbox-zero/blob/main/apps/web/app/(landing)/login/LoginForm.tsx): adds `showGoogleLogin` prop (defaults to `true` for backwards compat with existing callers and tests — Google was previously unconditional).
- `.env.example`: documents the new var next to the existing login env vars.
- Unit tests for the parser (`undefined`, empty, whitespace, single, multi, mixed-case, unknown tokens, all-unknown lockout guard).

## Out of scope

The "solve the bootstrap problem" Option A / B / C in the issue (CLI seed, first-run wizard, env-based SSO registration) is intentionally left out of this PR so the diff stays reviewable. This change lands the part that unblocks both the enterprise-lockdown and the brand-pruning use cases on its own; the bootstrap follow-ups can be addressed in dedicated PRs.

Refs #2106

### DIFF
--- a/apps/web/.env.example
+++ b/apps/web/.env.example
@@ -58,6 +58,7 @@ NEXT_PUBLIC_BYPASS_PREMIUM_CHECKS=true
 # AUTO_JOIN_ORGANIZATION_ENABLED=true # Self-hosted: auto-add new users to the org
 # SSO_LOGIN_ENABLED=true # Show the SSO button on the login screen
 # NEXT_PUBLIC_SHOW_APPLE_LOGIN=false # Show Apple login on the standard web login screen
+# NEXT_PUBLIC_LOGIN_PROVIDERS=google,microsoft,sso # Comma-separated list of login providers to show on the login screen. Unset = show all. Valid tokens: google, microsoft, apple, sso. Note: each provider must also be otherwise configured (e.g. Microsoft OAuth credentials, SSO_LOGIN_ENABLED, NEXT_PUBLIC_SHOW_APPLE_LOGIN) — this var can only narrow the visible buttons.
 # NEXT_PUBLIC_EXTERNAL_API_ENABLED=true # Enable the external API (v1 endpoints, API keys, API key UI)
 # NEXT_PUBLIC_WEBHOOK_ACTION_ENABLED=false # Self-hosted: disable outgoing webhook rule actions and webhook secret UI
 # AUTO_ENABLE_ORG_ANALYTICS=true # Self-hosted: default new org members to analytics enabled
--- a/apps/web/.env.example
+++ b/apps/web/.env.example
@@ -58,7 +58,7 @@ NEXT_PUBLIC_BYPASS_PREMIUM_CHECKS=true
 # AUTO_JOIN_ORGANIZATION_ENABLED=true # Self-hosted: auto-add new users to the org
 # SSO_LOGIN_ENABLED=true # Show the SSO button on the login screen
 # NEXT_PUBLIC_SHOW_APPLE_LOGIN=false # Show Apple login on the standard web login screen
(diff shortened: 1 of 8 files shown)


Ashut0sh-mishra:
Thank you @hex-security-app and @cubic-dev-ai -- you're 100% correct, and I'm glad the bots caught this before merge. The original PR description over-claimed the security properties: it was a UI-only filter, not real access control. Apologies.

Fixed in the latest commit. `apps/web/utils/auth.ts` now reads the same `NEXT_PUBLIC_LOGIN_PROVIDERS` allowlist server-side and gates **provider registration**, not just UI rendering:

- `googleSocialProvider`, `microsoftSocialProvider`, `appleSocialProvider` are set to `null` when their provider isn't in the allowlist, so they never enter the `socialProviders` object passed to `betterAuth(...)`.
- The OAuth-emulator entries in `genericOauthConfig` are gated the same way.
- The `sso` plugin is conditionally added to the plugins array based on the allowlist.

After this change, `POST /api/auth/sign-in/social` with an excluded provider returns 404 from better-auth itself -- the direct-API bypass is closed at the protocol level, not the UI.

The existing `getEnabledLoginProviders` helper is reused (no React/Next imports, just `@/env`), so parsing rules stay in one place.

`.env.example` updated to reflect the stronger guarantee. PR description / title will read the same -- this is now a backend access-control mechanism, not a cosmetic toggle.

Happy to add tests if you'd like (e.g. a unit test asserting `socialProviders` shape under different env-var values).

Ashut0sh-mishra:
Thanks @cubic-dev-ai -- added the wiring tests you asked for in apps/web/utils/auth.test.ts. They mock `getEnabledLoginProviders` per test and dynamically re-import `@/utils/auth` so the assertions run against a fresh `betterAuthConfig`:

- Excluding `google` / `microsoft` / `apple` from the allowlist -> the corresponding key is absent from `options.socialProviders`.
- Excluding `sso` -> the `sso` plugin is no longer in `options.plugins` (verified by comparing the plugin count with vs without sso).
- An sso-only allowlist drops every social provider.

This should catch any future refactor that accidentally re-exposes the OAuth endpoints without going through the allowlist. Pushed.

PR #2657, opened 2026-05-13
assistant: Pin follow-up replies to the source email

gentlemandev:
## Summary
- When a user replies to a follow-up notification in Slack/Telegram, the agent was sometimes drafting against the wrong email because it would run `searchInbox` and pick the most recent thread. We now persist the originating email reference at notification time and inject it as hidden chat context on reply, so the agent calls `replyEmail` with the correct `messageId` directly.
- The same context also disambiguates which connected email account to use when the chat is linked to multiple accounts.

## Changes
- Add `messaging-follow-up-context` Redis store keyed by `provider:channelId:messageTs` (14-day TTL).
- Save the email reference (account, thread, message, tracker, subject, counterparty) when sending Slack and Telegram follow-up notifications.
- On incoming Slack thread replies and Telegram replies (`reply_to_message_id`), look up the stored context, prefer the matching email account when candidates are ambiguous, and prepend a hidden user message that pins the referenced email for the model.
- `sendFollowUpReminderToSlack` now returns the resolved channel + message ts so we can key the context.

## Performance impact
- Adds 1 Redis `SET` per follow-up notification sent and 1 Redis `GET` per reply on a multi-candidate follow-up thread. Both are guarded behind `isRedisConfigured()` and only fire on the existing notification/reply path, so no impact on hot inbox routes.

## Test plan
- [x] `pnpm test apps/web/utils/redis/messaging-follow-up-context.test.ts`
- [x] `pnpm test apps/web/utils/follow-up/send-follow-up-notification.test.ts`
- [x] `pnpm test apps/web/utils/messaging/chat-sdk/bot.test.ts`
- [ ] Manual: trigger a follow-up reminder in Slack, reply in the thread asking "draft a reply", confirm the agent uses the correct `messageId`.
- [ ] Manual: same flow in Telegram with `reply_to_message` on the bot's notification.

### DIFF
--- a/apps/web/app/api/follow-up-reminders/process.ts
+++ b/apps/web/app/api/follow-up-reminders/process.ts
@@ -542,6 +542,11 @@ async function processFollowUpsForType({
               }) ?? undefined,
             threadLinkLabel: getThreadLinkLabel(providerName),
             trackerId: tracker.id,
+            emailReference: {
+              emailAccountId: emailAccount.id,
+              threadId: thread.id,
+              messageId: lastMessage.id,
+            },
             logger: threadLogger,
           });
         } catch (notifyError) {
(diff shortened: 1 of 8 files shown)


gentlemandev:
Closing as stale — 51 days with no activity since creation, now conflicting with main. The underlying bug (INB-257: agent drafts reply to wrong email from a follow-up notification thread) is still unfixed; this branch would need a fresh implementation rather than a rebase given how long it's diverged. Tracked in INB-257.

PR #2665, opened 2026-05-14
cold-email: write LEARNED_PATTERN matchMetadata on pattern hits

### LINKED ISSUE #2664: Bug: cold-email matchMetadata is always {type: "AI"}, even on pattern hits
## Summary

The cold-email blocker has two decision paths: ask the LLM, or check the learned-pattern list for a known sender. When the pattern path fires, no LLM call is made — but the resulting `ExecutedRule.matchMetadata` is still written as `[{type: "AI"}]`. The decision and its cost are correct; only the audit trail is wrong.

Every other rule type writes a `LEARNED_PATTERN` matchReason (with `group` and `groupItem`) on a pattern hit. The cold-email path is the only inconsistent one.

## Impact

Analytics that read `matchMetadata` (cost dashboards, "is pattern learning working?" reporting) overcount AI and undercount pattern hits on the cold-email path. On a 3-month self-hosted dataset I found 608 cold-email rows tagged `{type: "AI"}`; backfilling against `GroupItem` showed **~50% (302) were actually free pattern hits**.

No effect on cost or on which emails get blocked.

## Where

`apps/web/utils/ai/choose-rule/match-rules.ts` (~line 111, `main` @ `91a78dffd`):

```ts
return {
  matches: [
    { rule: coldRule, matchReasons: [{ type: ConditionType.AI }] },  // always "AI"
  ],
  ...
};
```

`isColdEmail()` in `apps/web/utils/cold-email/is-cold-email.ts` finds the matching `GroupItem` but discards everything except `exclude`, so the caller has no way to construct a `LearnedPatternMatch`.

## Reproduction

1. Enable cold-email blocker.
2. Trigger an LLM cold-email detection so a `GroupItem` (type `FROM`, `exclude: false`) is added.
3. Send a second email from the same sender.
4. Inspect `ExecutedRule.matchMetadata`.

**Expected:** `[{type: "LEARNED_PATTERN", group: {...}, groupItem: {...}}]`
**Actual:** `[{type: "AI"}]`

## Suggested fix

1. In `is-cold-email.ts`: expand the `groupItem.findFirst` select to include `id`, `type`, `value`, `exclude`, plus `group: { select: { id: true, name: true } }`. Return a `patternMatch: { group, groupItem }` field when `reason === "ai-already-labeled"`.
2. In `match-rules.ts`: when `coldEmailResult.patternMatch` is set, emit a `LearnedPatternMatch` matchReason instead of `{type: "AI"}`. Fall back to `{type: "AI"}` if the (nullable) group relation is missing.

Likely introduced by #1180 (Cold Email: refactor to use learned patterns) — the pattern lookup was added but the corresponding `matchReasons` construction was not. Happy to send a PR.

JasonBates:
Fixes #2664.

## Summary

Cold-email pattern hits are currently logged as `[{type: "AI"}]` in `ExecutedRule.matchMetadata` even when no LLM call was made (`isColdEmail` returned via a `GroupItem` lookup, reason `"ai-already-labeled"`). The decision and cost are correct — only the audit trail is wrong. Every other rule type emits a `LearnedPatternMatch` reason (with `group` and `groupItem`) on pattern hits via `serializeMatchReasons`; the cold-email path was the only inconsistent one.

The bug originated in #1180, which added the pattern lookup in `is-cold-email.ts` but didn't propagate the result back into the `matchReasons` shape constructed in `match-rules.ts`.

## Changes

- `utils/cold-email/is-cold-email.ts`: expand the `groupItem.findFirst` select to include `id`, `type`, `value`, `exclude`, and the `group: {id, name}` relation. Return a new optional `patternMatch: { group, groupItem }` field when `reason === "ai-already-labeled"`.
- `utils/ai/choose-rule/match-rules.ts`: when `coldEmailResult.patternMatch` is set, emit a `LearnedPatternMatch` matchReason instead of `{type: "AI"}`. Falls back to `{type: "AI"}` if the (nullable) Prisma `group` relation is missing.
- `utils/cold-email/is-cold-email.test.ts`: updated three test cases to assert the new select shape and the new `patternMatch` return value.

The emitted shape conforms to `serializedMatchReasonSchema` in `app/api/chat/validation.ts` and is already handled by the `LEARNED_PATTERN` case in `formatSerializedMatchMetadata` (`utils/ai/assistant/chat.ts`), which now renders cold-email pattern hits as e.g. `Cold Email learned pattern include on FROM: sender@example.com` instead of `Matched by AI instructions.`

## No behavior change

- No change to which emails get blocked.
- No change to LLM cost (the pattern path already skipped the LLM call — that part was correct).
- `shouldAnalyzeSenderPattern` in `run-rules.ts` already short-circuits on `SystemType.COLD_EMAIL` regardless of `matchReasons` type, so the type swap doesn't change pattern-analysis behavior.
- The new `patternMatch?` field is optional, so the other caller (`utils/actions/cold-email.ts` → `TestRules.tsx`) is unaffected.

## Test plan

- [x] `vitest run utils/cold-email/is-cold-email.test.ts utils/ai/choose-rule/match-rules.test.ts` → 111 passed
- [x] `tsc --noEmit` clean for the three changed files

### DIFF
--- a/apps/web/utils/ai/choose-rule/match-rules.ts
+++ b/apps/web/utils/ai/choose-rule/match-rules.ts
@@ -109,7 +109,15 @@ export async function findMatchingRules({
         matches: [
           {
             rule: coldRule,
-            matchReasons: [{ type: ConditionType.AI }],
+            matchReasons: coldEmailResult.patternMatch
+              ? [
+                  {
+                    type: ConditionType.LEARNED_PATTERN,
+                    group: coldEmailResult.patternMatch.group,
+                    groupItem: coldEmailResult.patternMatch.groupItem,
+                  },
+                ]
+              : [{ type: ConditionType.AI }],
           },
         ],
         reasoning: coldEmailResult.aiReason || coldEmailResult.reason,
--- a/apps/web/utils/cold-email/is-cold-email.test.ts
+++ b/apps/web/utils/cold-email/is-cold-email.test.ts
@@ -41,7 +41,10 @@ describe("isColdEmail", () => {
     // Mock groupItem lookup
     vi.mocked(prisma.groupItem.findFirst).mockResolvedValue({
       id: "group-item-id",
+      type: GroupItemType.FROM,
+      value: normalizedEmail,
       exclude: false,
+      group: { id: groupId, name: "Cold Email" },
     } as any);
 
     const email: EmailForLLM = {
@@ -62,6 +65,15 @@ describe("isColdEmail", () => {
 
     expect(result.isColdEmail).toBe(true);
     expect(result.reason).toBe("ai-already-labeled");
+    expect(result.patternMatch).toEqual({
+      group: { id: groupId, name: "Cold Email" },
+      groupItem: {
+        id: "group-item-id",
+        type: GroupItemType.FROM,
+        value: normalizedEmail,
+        exclude: false,
+      },
+    });
 
     // Verify that findFirst was called with the normalized email address
     expect(prisma.groupItem.findFirst).toHaveBeenCalledWith({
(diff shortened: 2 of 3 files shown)


elie222:
thanks!

PR #2667, opened 2026-05-14
Fix assistant inline markup rendering

gentlemandev:
Normalize streamed assistant inline markup before rendering so structured email cards do not appear as raw tags in list responses. Also unwrap block-style assistant components from markdown paragraphs to avoid invalid nesting.

- Normalizes assistant tag indentation, smart quotes, and line breaks before Streamdown parsing
- Unwraps block assistant components and block code from generated paragraphs
- Adds focused coverage for inline detail rendering inside numbered lists

Tests:
- `pnpm exec vitest run components/assistant-chat/assistant-inline-email-response.test.tsx` (fails in this environment before assertions with `React.act is not a function`)
- `pnpm --filter inbox-zero-ai test -- --run components/assistant-chat/assistant-inline-email-response.test.tsx` (mis-forwarded by package script and ran the broader suite; unrelated existing environment failures include `React.act is not a function` and missing `@/utils/hash`)

Performance impact: Adds small string normalization and child inspection only while rendering assistant responses. No new database or network calls; low hot-path risk because work is scoped to chat response rendering.

### DIFF
--- a/apps/web/components/assistant-chat/assistant-inline-email-response.test.tsx
+++ b/apps/web/components/assistant-chat/assistant-inline-email-response.test.tsx
@@ -9,7 +9,10 @@ import {
   waitFor,
 } from "@testing-library/react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import { AssistantInlineEmailResponse } from "@/components/assistant-chat/assistant-inline-email-response";
+import {
+  AssistantInlineEmailResponse,
+  normalizeAssistantInlineMarkup,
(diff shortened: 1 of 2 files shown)


gentlemandev:
Closing as stale — 50 days with no activity since creation, now conflicting with main. The underlying bug (INB-269: raw XML tags rendering in chat) is still unfixed; a partial fix landed separately via #2728, but the remaining cases this PR targeted are still broken. Needs a fresh implementation. Tracked in INB-269.

PR #2672, opened 2026-05-15
Fix inbox sender search filtering

gentlemandev:
Adds structured exact sender filtering for inbox search so Gmail and Outlook can search by sender address without relying on broad text syntax.

- Adds searchInbox.fromEmail and forwards it through email providers.
- Applies exact Outlook sender filtering through OData and paginated fallback paths.
- Updates provider guidance and targeted tests for sender search behavior.
- Tests: pnpm test utils/ai/assistant/chat-inbox-tools.test.ts utils/email/microsoft.test.ts


gentlemandev:
Closing as stale — 49 days with no activity since creation, now conflicting with main. The underlying bug (INB-272: incomplete Outlook sender search results) is still unfixed; this branch would need a fresh implementation rather than a rebase given how long it's diverged. Tracked in INB-272.

PR #2731, opened 2026-05-23
Fix messaging draft send state

gentlemandev:
Fixes the messaging notification send flow so an action that already owns a mailbox draft sends that existing draft and keeps the action state consistent.

- Uses the current notification action as the mailbox draft source when it already has a draft id.
- Avoids updating the same executed action twice after a draft send.
- Tests: `pnpm test utils/messaging/rule-notifications.test.ts`

### DIFF
--- a/apps/web/utils/messaging/rule-notifications.test.ts
+++ b/apps/web/utils/messaging/rule-notifications.test.ts
@@ -257,6 +257,7 @@ describe("handleRuleNotificationAction", () => {
 
     expect(provider.sendDraft).toHaveBeenCalledWith("draft-1");
     expect(provider.sendEmailWithHtml).not.toHaveBeenCalled();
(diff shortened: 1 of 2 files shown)


gentlemandev:
Re: cubic's ⚠️ on explicit verification of Gmail draft removal — addressed in 3fd6ff077.

Gmail's `drafts.send` API sends the draft and removes it from the drafts list as a single server-side operation; there is no separate delete call from our code. The existing assertions cover this contract:

- `provider.sendDraft` is called with the draft id (i.e. the Gmail-side delete-on-send path runs)
- `provider.sendEmailWithHtml` is not called (we don't bypass that path and leave the draft orphaned)

I added `expect(prisma.executedAction.update).toHaveBeenCalledTimes(1)` so the test now also locks in the second part of the fix (no duplicate `executedAction` update when `mailboxDraftAction.id === context.id`).

PR #2781, opened 2026-06-02
Fix assistant reply draft reuse

gentlemandev:
Fixes assistant chat reply confirmation so an existing rule-generated draft for the source message is reused instead of creating a separate reply. The confirmation flow now sends the existing draft and marks its action as likely sent.

- Reuses pending rule-generated reply drafts when preparing assistant chat replies.
- Carries reused draft metadata through validation and confirmation.
- Adds focused coverage for draft reuse during reply preparation and confirmation.
- Tests: `pnpm test utils/actions/assistant-chat.test.ts utils/ai/assistant/chat-inbox-tools.test.ts`
- Performance: adds one database lookup and, when a reusable draft exists, one provider draft fetch during reply preparation; confirmation may update/send the existing draft and mark its action status. Scope is limited to assistant reply preparation/confirmation, with low hot-path risk.

### DIFF
--- a/apps/web/utils/actions/assistant-chat-confirmation.ts
+++ b/apps/web/utils/actions/assistant-chat-confirmation.ts
@@ -1,6 +1,7 @@
 import "server-only";
 
 import type { Prisma } from "@/generated/prisma/client";
+import { DraftEmailStatus } from "@/generated/prisma/enums";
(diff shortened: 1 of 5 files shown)


elie222:
Closing as it gets messy

PR #2893, opened 2026-06-22
feat(webhooks): add WEBHOOK_ALLOW_PRIVATE_IPS opt-in for self-hosted internal targets

alexcf:
## Problem

Self-hosted users on a trusted network can't point a webhook rule action at their own internal services (LAN, Tailscale MagicDNS, etc.). The SSRF guards reject any webhook URL that **is**, or **resolves to**, a private IP with `Webhook URL cannot resolve to private IP addresses` (or `…cannot point to private IP addresses` for IP literals). Notably Tailscale's `100.64.0.0/10` CGNAT range is blocked, so MagicDNS targets like `https://host.tailnet.ts.net` are rejected.

This is the correct secure default for the hosted multi-tenant product, but a single-user self-host legitimately needs internal targets.

## Change

Adds a server-side env flag **`WEBHOOK_ALLOW_PRIVATE_IPS`** (`booleanString`, default `false`). When enabled, the private-IP checks are bypassed — and crucially in **both** guards a webhook send passes through:

1. **`apps/web/utils/webhook-validation.ts`** — `validateWebhookUrl()` (used at rule creation and again immediately before send). Skips the resolved-private-IP and private-IP-literal rejections.
2. **`apps/web/utils/network/safe-http-url.ts`** — `resolveSafeExternalHttpUrl()`, which the webhook sender (`webhook.ts`) uses to DNS-resolve the host and **pin the connection to that IP** (anti-rebinding). This is the guard that actually blocks a private-resolving *hostname* at send time; without gating it, a fix in `webhook-validation.ts` alone is insufficient. With the flag set, the resolved private address is allowed and pinned.

Behaviour is **byte-for-byte identical when the flag is unset** (the default). The flag deliberately does **not** relax `BLOCKED_HOSTNAMES` (localhost, named cloud-metadata endpoints) or HTTPS-only-in-production — these stay enforced as defense in depth, and relaxing them isn't needed to unblock internal HTTPS targets.

## Scope note

`safe-http-url.ts` is shared by three outbound paths: webhook delivery (`webhook.ts`), unsubscribe-link fetching (`senders/unsubscribe.ts`), and the upstash client (`upstash/index.ts`). The flag therefore also permits private-IP targets on those paths. For a trusted single-tenant self-host this is acceptable (and, for a self-hosted upstash/redis, arguably desirable); it is documented in code, the env description, and `.env.example` as an explicit, opt-in security trade-off. Happy to rename/rescope (e.g. a more general `ALLOW_PRIVATE_IP_TARGETS`) if you'd prefer.

## Security rationale

Enabling this disables SSRF protection for the app's outbound HTTP and must only be used on a trusted, single-tenant deployment. The flag is a non-public server var and defaults off, so hosted/multi-tenant deployments are unaffected unless an operator deliberately sets it.

## Validation

- Existing `webhook-validation.test.ts` and `safe-http-url.test.ts` pass unchanged with the flag unset.
- Added a `WEBHOOK_ALLOW_PRIVATE_IPS` suite to **both** test files: private IPv4/IPv6 literals and Tailscale CGNAT allowed via the literal and resolved paths; assertions that localhost / cloud-metadata hostnames and HTTPS-only-in-production are **still rejected** with the flag on; and regression checks that defaults reject as before.
- Verified end-to-end on a self-hosted deployment: with the flag set, a webhook rule action successfully delivered to a `https://*.ts.net` (Tailscale CGNAT) internal receiver that was previously rejected.

## Note on the helper

`allowPrivateIps()` is intentionally duplicated in both `webhook-validation.ts` and `safe-http-url.ts` (each reads `process.env` directly) to keep both modules dependency-free, matching the existing `NODE_ENV` reads in those files. Glad to extract a shared helper if preferred.

## Test plan

- Configure an internal/Tailscale webhook target; without the flag it's rejected, with `WEBHOOK_ALLOW_PRIVATE_IPS=true` it delivers.
- Confirm public targets and the default (flag unset) behaviour are unchanged.

<a href="https://cubic.dev/pr/elie222/inbox-zero/pull/2893?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


alexcf:
The terms look fine except this "If the Contributions include any method, process, or apparatus that is covered by a patent, the Contributor agrees to grant Inbox Zero a non-exclusive, worldwide, royalty-free license under any patent claims necessary to use, modify, distribute, and otherwise exploit the Contributions for the purposes of the Inbox Zero project." - I can't do that 😄 - if I don't own the patent, how can I provide you any kind of license, let alone royalty-free access to it?

alexcf:
Thanks both — addressed in 8806d98.

**@hex-security-app** — correct and important catch. The env-gated bypass in `safe-http-url.ts` also relaxed the unsubscribe path, which fetches attacker-controlled `List-Unsubscribe` URLs (genuine SSRF). Reworked to your option 1: `safe-http-url.ts` now takes an explicit `allowPrivateIps` option (default `false`); **only** the webhook sender passes it, so `unsubscribe.ts` and the upstash client keep private IPs blocked for untrusted input. Also applied option 3 — `metadata.google.internal` / `metadata.gcp.internal` are now blocked unconditionally as defense-in-depth.

**@cubic** — the `.env.example` scope concern is resolved by the same change: the flag is genuinely webhook-scoped again (no longer affects unsubscribe/upstash), so the comment is now accurate.

alexcf:
@cubic the scope flagged here changed in 8806d98: `safe-http-url.ts` no longer reads the env var — it takes a per-call `allowPrivateIps` option that **only the webhook sender passes**. The unsubscribe path and Upstash client call it without the option, so they stay restricted. The flag is therefore webhook-scoped, and applying the suggested wording (listing unsubscribe/Upstash) would be inaccurate. Ive clarified the `.env.example` comment to state the webhook-only scope explicitly.

PR #2898, opened 2026-06-23
feat(digest): add opt-in webhook delivery target for digests

alexcf:
Digests currently deliver only via email / Slack / Teams / Telegram. This adds a per-account **webhook** delivery target so self-hosters and integrators can receive digests programmatically.

When a `digestWebhookUrl` is set, each digest is POSTed as JSON `{ type: "digest", date, ruleNames, itemsByRule }`. A non-2xx response throws, so it counts as a failed channel under the existing `Promise.allSettled` handling (consistent with Slack) and satisfies the existing "≥1 deliverable channel" guard.

Includes: `digestWebhookUrl String?` on `EmailAccount` + migration, a validated server action, and a "Digest webhook URL" input in the digest settings UI.

**Opt-in:** the UI field is gated behind `NEXT_PUBLIC_DIGEST_WEBHOOK_ENABLED` (following the `NEXT_PUBLIC_DIGEST_ENABLED` pattern). With the flag off and no URL set, behavior is unchanged (nullable column, no delivery promise, hidden field).

**Validated:** `tsc --noEmit` clean on touched code, `biome` clean, `send-digest.test.ts` + `webhook.test.ts` pass (10). Migration generated under `prisma/migrations/`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<a href="https://cubic.dev/pr/elie222/inbox-zero/pull/2898?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #2901, opened 2026-06-23
feat(messaging): add Webhook channel provider for digest delivery

alexcf:
Supersedes #2898.

## Problem
Self-hosters cannot deliver digests to any chat channel without standing up an OAuth app (Slack) or bot token (Teams/Telegram). The account-level `digestWebhookUrl` in #2898 worked but lives outside the `MessagingChannel`/`MessagingRoute` framework and cannot be managed from the Channels page.

## Solution
Add `WEBHOOK` as a `MessagingProvider`. Users add a webhook on the **Channels** page with just a URL (+ optional secret) — no OAuth — and toggle it on for **Digests**. Digest sending POSTs `{ type: "digest", date, ruleNames, itemsByRule }` to the URL.

## SSRF-safe delivery
A new `sendDigestToWebhook` mirrors `utils/webhook.ts`: URL resolved via `resolveSafeExternalHttpUrl` and the request DNS-pinned to public addresses over `node:http`/`node:https` (no `fetch`). It **throws** on a blocked or non-2xx response so a failed webhook counts as a failed channel under `Promise.allSettled` and the digest is not marked SENT. The optional secret is sent as `X-Webhook-Secret`, stored **encrypted at rest** via prisma-extensions, and never returned by the API.

## Notes
- One webhook per account (`teamId=""`; create upserts so disconnect→re-add reconnects).
- Premium-gated: enabling Digests on a webhook calls `assertCanUseDigests`.
- Scoped to Digests; does not participate in rule notifications / meeting briefs / filings.
- Backwards compatible — new code only runs when a user adds a WEBHOOK channel.
- Two migrations: enum value (isolated, `ADD VALUE IF NOT EXISTS`) + `webhookUrl`/`webhookSecret` columns.

## Test plan
- `prisma generate`; `tsc --noEmit` clean on touched code; `biome` clean.
- Unit tests: `sendDigestToWebhook` (SSRF block, secret header, omitted secret, non-2xx throw); `sendDigest` WEBHOOK case (delivery, failure counts, non-operational skip); messaging-channels route (secret never returned, WEBHOOK available).
- Manual: add a webhook on Channels, enable Digests, trigger a digest, confirm POST + `X-Webhook-Secret`; confirm private/loopback URLs are rejected.
- Not yet run: migrations against a live DB.

<a href="https://cubic.dev/pr/elie222/inbox-zero/pull/2901?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/web/app/(app)/[emailAccountId]/channels/Channels.tsx
+++ b/apps/web/app/(app)/[emailAccountId]/channels/Channels.tsx
@@ -13,6 +13,7 @@ import {
   MoreVerticalIcon,
   Settings2Icon,
   SunIcon,
+  WebhookIcon,
 } from "lucide-react";
 import Image from "next/image";
 import { useAction } from "next-safe-action/hooks";
@@ -59,7 +60,11 @@ import {
   toggleRuleChannelAction,
   createMessagingLinkCodeAction,
   disconnectChannelAction,
+  createWebhookChannelAction,
+  updateWebhookChannelAction,
+  toggleWebhookDigestsAction,
 } from "@/utils/actions/messaging-channels";
+import { Input } from "@/components/ui/input";
 import { useSlackNotifications } from "@/app/(app)/[emailAccountId]/settings/ConnectedAppsSection";
 import { ProactiveUpdatesSetting } from "@/app/(app)/[emailAccountId]/assistant/settings/ProactiveUpdatesSetting";
 import { toastSuccess, toastError } from "@/components/Toast";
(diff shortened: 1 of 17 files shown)


alexcf:
Thanks @cubic and @hex-security-app — addressed the two security/correctness issues in the latest commit:

**1. Webhook secret silently erased on update (P1)** — confirmed: `webhookSecret` is `.optional()`, so leaving the field blank sends `undefined`, and `webhookSecret || null` overwrote the stored secret with `null`, breaking signature verification on later deliveries. Both `updateWebhookChannelAction` and the create-upsert `update` branch now only write `webhookSecret` when it's explicitly provided (conditional spread), preserving the existing value otherwise. The `create` branch still defaults to `null` for brand-new channels.

**2. Secret transmitted over plaintext HTTP** — `sendDigestToWebhook` now throws if a secret is set and the resolved URL isn't `https:`, so `X-Webhook-Secret` is never sent in the clear. Plain HTTP is still permitted when no secret is configured. (Enforced at send time so it holds regardless of how the channel was configured, including pre-existing rows.)

On the remaining points:
- **Duplicate HTTP transport vs `utils/webhook.ts`** — the duplication is deliberate (the file notes it mirrors `utils/webhook.ts`); the two have different throw-vs-swallow semantics. Extracting a shared SSRF-pinned transport is a reasonable follow-up but I've kept it out of this PR to avoid coupling two security-sensitive paths in one change.
- **Test coverage (vacuous secret-stripping assertion, WEBHOOK response case, POST-body assertion) and the shared create/update schema** — all fair. I've limited this commit to the verified source fixes; happy to follow up with the test/DRY improvements.

PR #2906, opened 2026-06-24
feat(assistant): sort the rules list A→Z and add inline search

alexcf:
### What
Two related improvements to the automation/assistant **Rules** table (`apps/web/app/(app)/[emailAccountId]/assistant/Rules.tsx`):

1. **Sort the rules strictly A→Z by name** (case-insensitive), instead of the previous enabled-first / system-grouped order — so a specific rule is always where you'd expect it.
2. **Inline search box** above the table that filters the visible rules by **name** and **condition text** (`conditionsToString`), case-insensitive, with a clear empty state.

### Why
As rules accumulate (system + many custom), locating one is slow: the grouped order is unpredictable when you're scanning for a name, and there's no filter. Alphabetical order + live search makes finding a rule immediate.

### How
- Display ordering: replaced `sortRulesForAutomation(...)` with a plain `name.localeCompare(..., { sensitivity: "base" })` sort over the combined system-placeholder + user rules list (purely presentational; no change to rule evaluation, which is server-side).
- Search: `searchQuery` state + a `filteredRules` `useMemo`; the table maps `filteredRules`; empty-state row when nothing matches.
- All client-side over data already fetched by `useRules()` — no API/query/server changes.
- Formatted with Biome; minimal diff (table body unchanged apart from `rules.map` → `filteredRules.map` + the empty row).

### DIFF
--- a/apps/web/app/(app)/[emailAccountId]/assistant/Rules.tsx
+++ b/apps/web/app/(app)/[emailAccountId]/assistant/Rules.tsx
@@ -10,8 +10,9 @@ import {
   Trash2Icon,
   SparklesIcon,
   CopyIcon,
+  SearchIcon,
 } from "lucide-react";
-import { useMemo } from "react";
+import { useMemo, useState } from "react";
 import { LoadingContent } from "@/components/LoadingContent";
 import { Button } from "@/components/ui/button";
 import { Card, CardDescription, CardHeader } from "@/components/ui/card";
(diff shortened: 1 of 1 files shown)


elie222:
Hey, we prefer current ordering as most people have just hte default rules set.

After the default ordering then alphabetical order is fine

alexcf:
Fixed! Default/system rules now stay in their canonical order first, then user-created rules are sorted alphabetically after them.

PR #2907, opened 2026-06-24
feat(assistant): auto-send the fix-rule message instead of only populating the input

alexcf:
### What
In the **Improve Rules** dialog (`FixWithChat`), after the user selects the expected rule (new / none / existing) and clicks **Next**, the generated message is now **sent automatically** to the assistant instead of merely being written into the chat input for the user to send manually.

### Why
The extra manual "press send" step is friction — the user already expressed full intent by choosing the rule and (optionally) an explanation. Auto-sending completes the action they started.

### How
- `FixWithChat` now pulls `submitTextMessage` and `chat` from `useChat()`.
- `handleSubmit` is `async`; after `setContext(...)` and opening the sidebar it calls `await submitTextMessage(input)`.
- `submitTextMessage(text)` takes the message as an **argument**, avoiding the `setInput` → submit state-closure race (the input would otherwise still be empty on the same tick).
- Guarded by `chat.status === "ready"`: if a response is already streaming/submitted, it **falls back to the previous behaviour** (`setInput(input)`) so we never interleave with an in-flight message.

### Scope
Only the `FixWithChat` ("Improve Rules") flow. The separate "Edit via AI" dropdown action (which intentionally lets you tweak the prompt before sending) is unchanged.

### Testing
Manual: select new/none/existing rule → Next → message is sent immediately; while a response is streaming, it populates the input instead.

<a href="https://cubic.dev/pr/elie222/inbox-zero/pull/2907?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/web/app/(app)/[emailAccountId]/assistant/FixWithChat.tsx
+++ b/apps/web/app/(app)/[emailAccountId]/assistant/FixWithChat.tsx
@@ -1,6 +1,6 @@
 import { MessageCircleIcon } from "lucide-react";
 import type { ChangeEvent } from "react";
-import { useMemo, useState } from "react";
+import { useMemo, useRef, useState } from "react";
 import { Button } from "@/components/ui/button";
 import type { ParsedMessage } from "@/utils/types";
 import type { RunRulesResult } from "@/utils/ai/choose-rule/run-rules";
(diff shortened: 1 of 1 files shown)


alexcf:
Thanks @cubic — both valid, fixed in the latest commit:

1. **Double-submit guard**: added an `isSubmittingRef` set at the top of `handleSubmit` (and cleared at the end), so repeated clicks in the async window before the modal closes are ignored.
2. **Send error handling**: `submitTextMessage` is now wrapped in try/catch — on failure it falls back to populating the chat input (so the generated text isn't lost) and reports via `captureException`. State cleanup still runs.

PR #2943, opened 2026-07-02
messaging: Add HTTPS webhook delivery for digests

elie222:
Scheduled digests can currently be delivered to email, Slack, Teams, or Telegram. This adds one webhook channel per email account so users can receive digest JSON at an HTTPS endpoint without configuring a chat app or bot.

Supersedes #2901 (which replaced #2898). Updated against current main, preserving the original contribution and resolving the delivery-settings and transactional-email test conflicts.

- Configure, reconnect, or remove the endpoint on Channels, and enable digests with Plus plan access.
- POST `{ type: "digest", date, ruleNames, itemsByRule }`. Items include sender, subject, and digest content; receivers must handle this custom payload.
- Require HTTPS at save time and send time, with or without a secret. Reject URL credentials; validate DNS and pin the outgoing connection. Honor the existing operator-controlled `WEBHOOK_ALLOW_PRIVATE_IPS` opt-in for trusted self-hosted networks.
- Encrypt the optional secret at rest and send it as `X-Webhook-Secret`. API responses expose only a presence flag. Blank edits preserve it, and an explicit control removes it. Whitespace-only secrets are never sent.
- Keep webhook delivery scoped to digests, excluding unsupported meeting-brief and other delivery settings.

Delivery retains existing multi-channel semantics: the overall digest succeeds if any destination succeeds. A failed webhook is logged, but is not independently retried after another destination succeeds. This is also stated in the Channels UI.

Validation:
- 58 focused unit tests passed: configuration, plan gating, account scoping, URL policy, sender, digest partial failures, API secret exclusion, and connection validity.
- Emulated Channels browser suite passed (authentication setup plus two product scenarios); inspected the configured-secret and reconnect screenshots.
- All Prisma migrations applied successfully to a disposable local PostgreSQL database.
- Biome and diff whitespace checks passed. Full build/type validation is delegated to CI.

The existing CLA status still needs to be cleared before merge. No merge is performed by this update.


elie222:
Maybe we'll add this in the future

PR #3062, opened 2026-07-27
Say what the confidence settings actually do, and record two dead ends

elie222:
No behaviour change. Copy fixes plus two comments recording experiments that failed, placed where someone would otherwise repeat them.


elie222:
Closing — this is almost entirely comments, and internal learnings belong in the private evals repo rather than here.

The two user-facing copy strings for the draft confidence settings are the only part that is a real product change; I will raise those separately if wanted, rather than smuggle them in behind a wall of commentary.

Noted for anything that follows: a change to how drafts are generated needs `DRAFT_PIPELINE_VERSION` in `apps/web/utils/ai/reply/draft-attribution.ts` bumped, so production telemetry stays segmentable by pipeline.

PR #3173, opened 2026-08-07
docs: Remove stale internal documentation

elie222:
Removes internal feature documentation that duplicated implementation behavior and had drifted from the code.

- Deletes obsolete architecture and feature notes.
- Updates contributor guidance to treat code and tests as the source of truth for behavior.
- Keeps operational and test documentation unchanged.

Validation: git diff --check.


elie222:
Regarding https://github.com/elie222/inbox-zero/pull/3173#pullrequestreview-4879279195: the E2E README findings describe existing base-branch content. Both files are byte-for-byte unchanged in the final PR diff; they appear as additions only in the follow-up commit that restored them after the scope was narrowed. Keeping those operational runbooks unchanged and out of this feature-document cleanup.

PR #3333, opened 2026-08-18
fix(docker): add NEXT_PUBLIC_EXTERNAL_API_ENABLED build placeholder

MGBuilds9:
## Problem

Self-hosters who set `NEXT_PUBLIC_EXTERNAL_API_ENABLED=true` get a working `/api/v1` server-side, but the **API Keys section never appears in Settings** — only the Webhook section shows.

Root cause: `docker/scripts/replace-next-public-placeholders.sh` can only substitute `NEXT_PUBLIC_*` values that were baked into the build output as `*_PLACEHOLDER` strings via the ARG/ENV pairs in `Dockerfile.prod`. `NEXT_PUBLIC_EXTERNAL_API_ENABLED` is missing from that list, so the client bundle keeps the build-time default (`false`) regardless of the runtime env. Since the server routes work, the missing UI reads as user error.

## Fix

Add the ARG/ENV placeholder pair, mirroring the adjacent `NEXT_PUBLIC_WEBHOOK_ACTION_ENABLED` entry. The placeholder matches the substitution script's regex (`NEXT_PUBLIC_[A-Z0-9_]+_PLACEHOLDER`), and `booleanString` coerces the substituted runtime value exactly as it does for the other flags (empty → default false).

Verified against a live self-hosted deployment: current image has no `NEXT_PUBLIC_EXTERNAL_API_ENABLED_PLACEHOLDER` occurrences in `.next/static`, so substitution has nothing to rewrite.

Note: a few other self-host-relevant flags (e.g. `NEXT_PUBLIC_BOOKING_LINKS_ENABLED`, `NEXT_PUBLIC_CONTACTS_ENABLED`, `NEXT_PUBLIC_FOLLOW_UP_REMINDERS_ENABLED`) are also absent from the placeholder list; happy to extend this PR if you want them included.

<a href="https://cubic.dev/pr/elie222/inbox-zero/pull/3333?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MGBuilds9:
Adding verification evidence, since this change is easy to eyeball but hard to trust without seeing the mechanism proven.

### Confirmation from a live self-hosted deployment

The two flags side by side in a running `web` container built from the current image, after `replace-next-public-placeholders.sh` has run at startup:

```
NEXT_PUBLIC_WEBHOOK_ACTION_ENABLED:""                                        <- has a placeholder pair, substituted
NEXT_PUBLIC_EXTERNAL_API_ENABLED:E.default.env.NEXT_PUBLIC_EXTERNAL_API_ENABLED   <- no pair, never inlined
```

The second one is never a string literal, so the startup script has nothing to rewrite and `booleanString` falls through to `.default(false)` permanently. `grep -ro "NEXT_PUBLIC_EXTERNAL_API_ENABLED_PLACEHOLDER" apps/web/.next/static` returns nothing in the shipped image, confirming it.

The server half is unaffected, which is what makes this confusing to diagnose: `/api/v1/openapi` returns 200 and `/api/v1/rules` correctly returns `401 Missing API key` while the Settings UI shows only the Webhook section. It reads like operator error rather than a packaging gap.

### Test matrix

I reproduced the exact `experimental__runtimeEnv` pattern in a minimal Next 15.5 app: one flag with an ARG/ENV placeholder pair, one without. Build output matched the production image exactly (literal string vs. un-inlined `process.env` reference).

I then ran the unmodified upstream `replace-next-public-placeholders.sh` + `replace-placeholder.sh` against that real build output inside `node:24-alpine`, the same base as the runner stage:

| runtime env | resulting bundle value | feature |
|---|---|---|
| `=true` | `"true"` | on |
| `=false` | `"false"` | off |
| unset | `""` | off (default preserved, no regression) |

Also verified: no placeholder tokens left behind, no `sed` backup artifacts, every emitted chunk still parses with `node --check`, and re-running the script (container restart) is idempotent. Values containing `|`, `&` and backslashes are escaped correctly by the existing `ESCAPED_VALUE` logic.

### Two things I checked that could have made this unsafe, and did not

**Build-time truthiness.** Inside the builder stage the flag holds the literal placeholder string, which `booleanString` evaluates to `true`. If a statically prerendered route gated on this flag, "enabled" markup could be frozen into the HTML. Checked `prerender-manifest.json` on the built image: 16 static routes (`/`, `/login/*`, `/home`, …) and none of them are settings or API-key routes, which are dynamic behind auth. No exposure. This is also inherent to the existing 11 pairs, not something this PR introduces.

**Placeholder collisions.** No other `NEXT_PUBLIC_*` name is a substring of `NEXT_PUBLIC_EXTERNAL_API_ENABLED`, and the greedy `[A-Z0-9_]+_PLACEHOLDER` match plus `sort -u` handles the set safely.

`docker build --check` on the patched Dockerfile reports no new findings; the only warnings are the pre-existing `SecretsUsedInArgOrEnv` ones on the dummy build-only secrets around lines 79-80.

### Scope

I kept this to the one flag that is actually broken in the wild. For context, `apps/web/env.ts` declares 60 `NEXT_PUBLIC_*` variables and `Dockerfile.prod` now carries 12 placeholder pairs, so 48 remain runtime-inert in Docker deploys. Most are Stripe/PostHog/analytics keys where nobody expects Docker runtime config, but a few look like genuine self-host feature gates with the same symptom, for example `NEXT_PUBLIC_BOOKING_LINKS_ENABLED`, `NEXT_PUBLIC_CONTACTS_ENABLED` and `NEXT_PUBLIC_FOLLOW_UP_REMINDERS_ENABLED` (verified absent).

Happy to extend this PR to cover those, or to add a build-time assertion that every client-side flag in `env.ts` has a matching placeholder so this cannot silently regress. Just say which you would prefer, or leave it as the single-flag fix if you would rather keep the diff minimal.

The one thing I have not done is a full `docker build` of the production image end to end. My change is structurally identical to the 11 adjacent pairs and BuildKit lints clean, but I wanted to be explicit about what was and was not exercised.

PR #3406, opened 2026-08-26
INB-337: Enable contact autocomplete for Cc and Bcc

gentlemandev:
> ✅ **Browser QA passed** · [QA report](https://qa.getinboxzero.com/20260827-225950_pr-3406_cf18771a6e5b/report.html)
>
> <sub>Apply `rerun-qa` to rerun.</sub>

Extends the composer contact-recipient control to To, Cc, and Bcc, while preserving valid pending recipient text during send. Recipient resolution is centralized so all three fields behave consistently.

- Reuse contact suggestions and reconnect handling across To, Cc, and Bcc.
- Commit and deduplicate pending valid input from every recipient field.
- Validate with 5 focused unit tests and Ultracite checks on all changed files.
- Performance: adds small client-side work per mounted recipient field; no database changes or new endpoints, while Cc/Bcc typing uses the existing contacts API through SWR. Hot-path risk is low and limited to the compose UI.

Linear: https://linear.app/inbox-zero-ai/issue/INB-337/contact-autocomplete-not-working-in-cc-and-bcc-fields-in-mail-composer

<a href="https://cubic.dev/pr/elie222/inbox-zero/pull/3406?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved To, Cc, and Bcc recipient entry with clearer selection, editing, and removal controls.
  * Added more reliable handling of pending recipients and incomplete email addresses.
  * Improved Cc/Bcc visibility controls and restored focus when fields are shown again.

* **Bug Fixes**
  * Improved attachment and email mutation identifier generation, including support for environments without `crypto.randomUUID()`.

* **Tests**
  * Added coverage for recipient selection, pending recipients, and UUID generation.


gentlemandev:
Browser QA failed with `crypto.randomUUID is not a function` before the mail page rendered, so the Cc/Bcc autocomplete flow was never exercised. Root cause: the QA browser runs over plain HTTP (insecure context), where `crypto.randomUUID` is unavailable — the crash came from the durable outbox manager that mounts on the mail page (pre-existing on main, not this PR's changes), but it blocks QA here and would also break self-hosted HTTP deployments.

Fixed in 4e549080e: added `utils/uuid.ts` `randomUuid()` (falls back to `crypto.getRandomValues`, which works in insecure contexts) and switched all client-side `crypto.randomUUID` call sites to it (outbox manager, thread actions, email-cache mutations, composer attachments, assistant chat, booking page). Server-side call sites are unchanged since Node always provides `randomUUID`. Covered by `utils/uuid.test.ts`.

gentlemandev:
@coderabbitai Addressed your review findings in 0c33d6404:

- **Removal of the last selected recipient (ComposeEmailForm.tsx onChange):** Confirmed — deselecting the sole recipient passed `[]` and the early return kept the stale value. Extracted the selection logic into `resolveRecipientSelection()` in `compose-recipients.ts`, which resolves an empty array to an empty selection, and added regression tests in `compose-recipients.test.ts`.
- **Keyboard-accessible recipient badges:** The click-to-edit handler moved off the `Badge` onto a dedicated `button type="button"` (with an `Edit` aria-label) alongside the existing remove button, so both actions are focusable and keyboard-operable. This also removed the need for `stopPropagation` on the remove button.
- **`contact.emailAddress[0]` → `.at(0)`:** Applied.

The docstring-coverage pre-merge warning is out of scope: this repo's convention is self-documenting code with why-only comments, not docstrings.

elie222:
Regarding the accessibility note in https://github.com/elie222/inbox-zero/pull/3406#pullrequestreview-5046213207: verified that the Cc/Bcc focus handlers and behavior are unchanged from the current main branch, so this is outside this PR's diff and scope.

PR #3506, opened 2026-09-04
meetings: Open the related draft from Go to Drafts (INB-328)

bahmd:
## Why the previous attempts didn't work

Both #3328 and #3346 assembled a deeplink by hand from an id the target interface does not accept, which is why the button lands on a folder rather than the draft.

**Outlook.** The `deeplink/compose?itemid=` form expects an **EWS item id**, and we hold a **Graph REST id**. Those are different id spaces, so OWA cannot resolve the item and falls back to the mailbox. Graph already returns a [`webLink`](https://learn.microsoft.com/en-us/answers/questions/4619414/open-a-draft-message-created-with-ms-graph-api-for) that resolves the item with no translation at all, and `getDraftMessage` issues no `$select`, so the full default payload — `webLink` included — is already on every draft we fetch. We were building a link while holding the provider's own.

**Gmail.** This one cannot be fully fixed. The `compose=` parameter takes an internal id that is not derivable from an API id — [Google's own community thread](https://support.google.com/mail/thread/86596268/the-url-https-mail-google-com-mail-u-0-drafts-compose-draftmailid-doesn-t-work-anymore?hl=en) confirms the old `#drafts?compose=<messageId>` form stopped working, and a [follow-up discussion](https://community.latenode.com/t/navigating-to-draft-messages-in-gmails-updated-interface-using-api-generated-ids/11481) concludes the new URLs "use an internal encoding system that's not publicly accessible." So `#inbox?compose=` (#3328) could never have opened the composer, and neither could `#drafts/<messageId>` (#3346) — that one also passed a *message* id where Gmail conversation URLs take a *thread* id.

## Changes

- Pass the draft message to `getEmailDraftUrl` instead of an id, because the fields that yield a working link differ by provider.
- Outlook: use the Graph `webLink`, keeping the constructed URL as a fallback. Since this value is redirected to, it is only trusted once its host is confirmed to be an Outlook host over https — otherwise the endpoint would forward wherever the payload pointed.
- Gmail: link to the draft's conversation under Drafts, addressed by **thread** id, falling back to the message id when no thread is present.

## Result

Outlook opens the draft itself. Gmail opens the draft's conversation in Drafts, which is as close as its interface allows — the composer is not reachable from an API id.

## Validation

- 6 new tests: the provider link taking precedence, three rejected-link cases (wrong host, non-https, unparseable), and Gmail thread-vs-message id selection.
- Route tests cover both providers end to end.
- Full `apps/web` suite run before and after; the failure set is **identical** (8 files, 19 tests, all pre-existing), with 6 net-new passing tests.

## Note for reviewers

The Gmail half is the honest best available rather than a fix, and it is worth confirming on a real account which provider the reporter was using — if it was Outlook, this closes the ticket; if Gmail, the button's promise may need rewording instead, since the composer cannot be deeplinked.

<a href="https://cubic.dev/pr/elie222/inbox-zero/pull/3506?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved meeting follow-up draft redirects for Gmail and Outlook.
  - Gmail drafts now open directly in the relevant conversation within the Drafts folder.
  - Outlook drafts now use the provider’s supplied web link when trusted, with a fallback link when needed.
  - Updated draft-link handling to provide more reliable navigation across supported email providers.

### DIFF
--- a/apps/web/app/api/user/meeting-recorder/meetings/[meetingId]/draft/route.test.ts
+++ b/apps/web/app/api/user/meeting-recorder/meetings/[meetingId]/draft/route.test.ts
@@ -42,12 +42,15 @@ describe("meeting follow-up draft route", () => {
     getEmailAccountMock.mockResolvedValue("user@gmail.com");
   });
 
-  it("opens the related Gmail draft using its embedded message ID", async () => {
+  it("opens the related Gmail draft's conversation in Drafts", async () => {
     prisma.meeting.findFirst.mockResolvedValue({
       followUpDraftId: "draft-resource-123",
       emailAccount: { account: { provider: "google" } },
     } as never);
-    emailProvider.getDraft.mockResolvedValue({ id: "draft-message-123" });
+    emailProvider.getDraft.mockResolvedValue({
+      id: "draft-message-123",
+      threadId: "thread-123",
+    });
 
     const response = await GET(new NextRequest(requestUrl), routeContext);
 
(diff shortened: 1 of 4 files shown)


elie222:
Superseded by https://github.com/elie222/inbox-zero/pull/3640 (rebased + Cubic P2/P3 fixes; this fork head wasn’t writable). Closing in favor of the merged follow-up.

elie222:
Closing as superseded by #3640.

PR #3514, opened 2026-09-05
drive: Preserve existing files during OneDrive uploads

elie222:
> ⚪ **Browser QA skipped** · [QA report](https://qa.getinboxzero.com/20260906-084650_pr-3514_61baed00cbd7/report.html)
>
> <sub>Apply `rerun-qa` to rerun.</sub>

Small OneDrive attachment uploads used a filename-addressed PUT with the default replacement behavior, so filing another `invoice.pdf` into the same folder could overwrite an existing document. Request rename-on-conflict for simple uploads, matching the protection already used by upload sessions, and return the new file metadata.

- Regression tests cover a small file and the inclusive 4 MiB simple-upload boundary. Both failed before the fix because the conflict policy was absent; all 11 provider tests pass after the fix.
- Validation: `pnpm test utils/drive/providers/microsoft.test.ts`, focused Biome checks, and `git diff --check`.
- Same upload request and payload; no additional database calls or network requests. Tests mock Graph; no live account was modified.

Microsoft documents that PUT defaults to replacing conflicts and that the conflict policy belongs in the request URL: [driveItem conflict behavior](https://learn.microsoft.com/en-us/graph/api/resources/driveitem?view=graph-rest-1.0#instance-attributes).

<a href="https://cubic.dev/pr/elie222/inbox-zero/pull/3514?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * OneDrive uploads now automatically rename new files when a file with the same name already exists, preventing conflicts while preserving the uploaded content.
  * Simple uploads now handle both small and larger files consistently with resumable uploads.

### DIFF
--- a/apps/web/utils/drive/providers/microsoft.test.ts
+++ b/apps/web/utils/drive/providers/microsoft.test.ts
@@ -68,7 +68,8 @@ describe("OneDriveProvider", () => {
       size: content.length,
     }));
     const header = vi.fn(() => ({ put }));
-    const api = vi.fn(() => ({ header }));
+    const query = vi.fn(() => ({ header }));
+    const api = vi.fn(() => ({ header, query }));
 
     vi.mocked(Client.init).mockReturnValue({ api } as any);
 
@@ -98,7 +99,8 @@ describe("OneDriveProvider", () => {
       size: content.length,
     }));
     const header = vi.fn(() => ({ put }));
-    const api = vi.fn(() => ({ header }));
+    const query = vi.fn(() => ({ header }));
+    const api = vi.fn(() => ({ header, query }));
 
     vi.mocked(Client.init).mockReturnValue({ api } as any);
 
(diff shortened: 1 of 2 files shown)


elie222:
Addressed both suggestions from https://github.com/elie222/inbox-zero/pull/3514#pullrequestreview-5122535537 in 61baed00c: the new client mock uses an explicit Client type instead of any, and the new buffer assertion uses at() for both indices. All 11 OneDrive provider tests and focused Biome checks pass.

PR #3584, opened 2026-09-07
feat: learn sender from user-applied labels

jacob-heinz:
## Summary

Adds an opt-in **Learn from labels** setting (Assistant → Settings → Advanced). When it is on and the user moves an email to a label in Gmail, the sender is saved as a learned pattern for the rule that applies that label — so the next email from them gets the same label without an AI call. If no rule uses that label yet, one is created from the label (LABEL, plus ARCHIVE when the email was moved out of the inbox rather than only tagged).

This is the "train by filing" behaviour people know from SaneBox: drag an email to a folder once, and that sender lands there from then on.

## How it works

- `EmailAccount.learnFromLabels` (default `false`) + migration.
- `handleLabelAddedEvent` (Gmail webhook) already receives user label adds. With the setting on it now calls `learnSenderFromLabel` after the existing classification-feedback step. The existing self-labeling filter still applies, so labels Inbox Zero applied itself are never learned.
- `utils/rule/learn-from-label.ts`:
  - existing rule for the label → `saveLearnedPattern` (FROM sender, `source: LABEL_ADDED`).
  - no rule → create one named after the label via `createRuleWithResolvedActions`, then learn the sender. Skips if a rule with that name already exists but does not label with it.
- Removing the label still goes through `recordLabelRemovalLearning` (exclusion pattern), so moving an email back to the inbox undoes the training for that sender.
- Settings toggle `LearnFromLabelsSetting` + `enableLearnFromLabelsAction`, mirroring `MultiRuleSetting`.

Off by default; no behaviour change for existing accounts.

## Scope / not included

- Gmail only (this is in the Google webhook path). Outlook could follow the same shape.
- Learning is per sender address, matching how learned patterns already work.

## Test plan

- [x] `utils/rule/learn-from-label.test.ts` — existing rule, label-only rule, label-and-archive rule, unreadable label, name collision, creation failure.
- [x] `process-label-added-event.test.ts` — setting off, system labels, no rule, existing rule, system-applied label.
- [x] `biome check` clean on touched files; `tsc --noEmit` shows no errors in touched files.
- [ ] Manual: enable the setting, move an email to a label, confirm the sender appears under Learned patterns and the next email from them is labeled.

<a href="https://cubic.dev/pr/elie222/inbox-zero/pull/3584?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added an Advanced setting for Google accounts to enable or disable learning from manually applied labels.
  - When enabled, the app can learn sender-to-label rules based on labels users apply and whether messages remain in the inbox or are archived.
  - Removing a learned label can remove the sender from the corresponding rule when the message returns to the inbox.
  - Safeguards prevent learning from system-applied labels and conflicting data.
  - Newly learned sender patterns take precedence over previous rule assignments.
- **Tests**
  - Added coverage for label-based learning, undo behavior, and settings scenarios.


bibager:
Addressed the review in de505e7:

- **Undo on label removal** (cubic P1): removing a learned label now writes an exclusion for the sender on custom and label-created rules (`unlearnSenderFromLabel`), mirroring what already happens for learnable system rules. Moving an email back to the inbox untrains the sender.
- **Message read failure** (CodeRabbit): rule creation now aborts instead of assuming "still in inbox"; regression test added.
- **Concurrent creation** (CodeRabbit): on a unique-name conflict the rule another event just created is loaded and learned into; test added.
- **System-applied check with a disabled rule** (CodeRabbit): `wasLabelAppliedBySystem` now runs whenever learning is on, even if `findRuleByLabelId` returns nothing; test added. When learning is off and no rule matches, the extra query is still skipped.
- **Gmail-only** (CodeRabbit): the setting renders only for Google accounts and `enableLearnFromLabelsAction` rejects other providers.
- **Toggle**: `ariaLabel` added; disabled while the action is executing (cubic P2s).
- **Typed fixtures**: the provider mock is now typed as `EmailProvider`; Prisma row mocks keep the `as any` convention the surrounding tests use.

Not changed — **learning into `TO_REPLY` / `AWAITING_REPLY` / `FYI` / `ACTIONED`** (cubic P1 on line 254): those are per-email states, not sender categories, which is why their configs have `shouldLearn: false`. Training "everything from this sender is To Reply" from one labeled email would be wrong, so the early return stays for non-learnable system rules.

PR #3707, opened 2026-09-15
mail: Keep forwards in their thread and explain unreadable threads

elie222:
Forwarding from the reader sent no thread id, so the provider started a new conversation holding only the forward. The reader then opened that new thread before the provider could serve it, and the whole pane was replaced by the generic "There was an error / An unexpected error occurred" screen until a later revalidation happened to succeed.

- Compose now sends the thread id even when there is no message to reply to, which is the case for a forward. The guard that dropped the entire threading payload without a reply target had been in place since the compose metadata refactor
- The thread detail route maps a provider "not found" to a known 404 with an explanation, instead of an unhandled error that returns a 500 and reports to Sentry twice
- The reader shows that reason plus a retry, instead of the app-wide error screen
- Restoring a queued send from the outbox tells a forward from a reply by its reply target, since a forward now carries thread metadata too

The reader still follows a sent message into a new provider thread when one is genuinely created. For a forward that no longer happens, so the emulated spec that asserted the jump now asserts the opposite: the reader stays on the source thread, with the forwarded message still above.

Tests, each confirmed failing before the fix and passing after:

- `reply-to-email-payload.test.ts`: a forward keeps its thread id, a reply keeps its full threading metadata, a new email sends none
- `route.test.ts`: a provider "not found" becomes a known 404
- `ThreadReader.test.tsx`: a failed thread load explains itself and retries
- `reply-drafts.test.ts`: a queued forward restores as a forward
- `compose-and-reply.spec.ts`: a sent forward stays in the thread it came from, verified against the emulator locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conversation loading errors with clearer messages and a **Try again** option.
  * Thread-not-found errors now provide a helpful 404 response when a conversation is still syncing or unavailable.
  * Forwarded emails now preserve their originating conversation and restore correctly, even without a specific message reference.
  * Temporary loading errors are handled more reliably without prematurely displaying an error state.
* **Tests**
  * Added coverage for reply metadata, forwarding behavior, retry actions, and unavailable conversations.


elie222:
Thanks @coderabbitai. Handled as follows:

**Dev-only 404 suppression hiding the reader's error state** — valid, fixed in b5f446fd0. `LoadingContent` ignored every 404 in development so the reader spun forever instead of showing the message and the retry. It now keeps ignoring unmarked 404s, which is what the reloading dev server produces, and renders the ones the API marked `isKnownError`. Covered by a new `LoadingContent.test.tsx`, confirmed failing before the change.

**Outlook forwards not preserving the conversation** — accurate, but out of scope here and not a regression from this PR. The Graph send path has always ignored the thread id, so Outlook forwards started a new conversation before this change too; nothing in this diff makes that worse. Preserving it needs `createForward` against a resolved source message, which is a separate change.

**Unused `_description` parameter in the table test** — declining. The parameter is positional: it feeds `%s` in the test name and the following argument cannot be reached without it. Same pattern as the existing table test in `app/api/threads/[id]/route.test.ts`, and the linter is happy with the underscore prefix.

elie222:
@coderabbitai re https://github.com/elie222/inbox-zero/pull/3707#issuecomment-5683422572 — no follow-up issue for now, thanks. The Outlook `createForward` gap is noted and the repo owner will decide whether to track it.

PR #3714, opened 2026-09-15
mail: Show the files a forward carries

elie222:
A forward sends the files of the message it quotes, but the composer listed only the files the user added, so nothing on screen said the attachments were going along.

- List the forwarded message's files alongside the composer's own, using the metadata the reader already holds, so the bytes stay on the provider until the send
- Works the same on both providers: Gmail fetches the parts at send time, Outlook copies them onto the forward draft

Tests, confirmed failing before the change and passing after:

- `EmailMessage.test.tsx`: the forward composer receives the source message's files
- `compose-and-reply.spec.ts`: forwarding the seeded message that carries an attachment lists it in the composer, verified against the emulator locally

Not included: a way to drop one of these files before sending. Gmail can skip named parts, but Graph copies attachments onto the forward draft with fresh ids, so removal there means listing the copies and matching by name and size. That belongs in its own change rather than inside a visibility fix.

<a href="https://cubic.dev/pr/elie222/inbox-zero/pull/3714?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Forwarded emails now display the original message’s attachments in the forwarding composer.
  * Attachment details include the filename and file size, with a read-only indicator showing they came from the forwarded message.
* **Tests**
  * Added coverage verifying forwarded attachments appear in the composer without requiring downloads.
  * Added end-to-end coverage for forwarding messages with attachments.


elie222:
Taken, in the follow-up commit: `forwardedAttachments` is now `Pick<EmailAttachmentMetadata, "id" | "filename" | "mimeType" | "size">[]` and `ForwardedAttachment` is gone. It was also an exported type used only in its own file, which the repository guidelines call out, so the shared contract fixes both.

PR #3773, opened 2026-09-17
perf: Paint the homepage hero immediately and cut boot work

elie222:
## Summary

The homepage felt slow because the hero stayed hidden until Framer Motion hydrated, then the logged-in shell duplicated `/api/user/email-accounts` and over-fetched images and footer RSC payloads. This paints LCP content in the first HTML, keeps a CSS entrance animation, coalesces the account fetch, and drops the boot work that was not earning its keep.

- Hero headline, subtitle, CTA, logos, and video thumbnail are visible immediately (opacity stays 1). Motion is CSS: words still stagger/unblur, the rest still rise in the same delay sequence.
- Cap `next/image` `deviceSizes` at 1920 so retina desktops no longer download a 3840w hero variant.
- Disable marketing footer `Link` prefetch so the homepage does not fire dozens of RSC requests on load.
- Coalesce concurrent `/api/user/email-accounts` callers onto one in-flight request.
- Sample Sentry traces at 10% (errors and session replay rates are unchanged).

## UX tradeoffs

- **Hero animation is CSS, not Framer.** Same stagger and timing, but words never start at `opacity: 0`. They rise/unblur while remaining readable, so first paint is not a blank hero. `prefers-reduced-motion` skips the motion.
- **Not identical to the old effect.** Framer was hiding each word until JS. Now you can see the headline (slightly soft) while it sharpens. That is the speed/animation middle ground.
- **Hero image may look slightly softer on 3x displays.** Max srcset width is 1920 instead of 3840. 1x/2x and the `sizes` hint still cover typical viewports.
- **Footer links are a bit slower on first click.** They no longer prefetch in the background; the homepage stops issuing ~60 extra RSC requests.
- **Sentry performance traces are sampled.** About 90% of transactions will no longer show up. Errors still capture; session replay is unchanged.
- **GTM stays on the app shell.** Moving it would drop `CompleteRegistration` from onboarding. That is a follow-up, not this PR.

## Test plan

- [ ] Logged-out homepage: headline is readable before JS; words still stagger/unblur
- [ ] Subtitle, CTA, logos, and video thumbnail are visible immediately and still rise in sequence
- [ ] `prefers-reduced-motion: reduce` shows a static hero
- [ ] Marketing pages that use `WordReveal` still wrap correctly (including emphasized words)
- [ ] Hero thumbnail is not requested at 2048/2560/3840
- [ ] Homepage network tab does not prefetch every footer route
- [ ] Logged-in boot: `/api/user/email-accounts` fires once, not twice
- [ ] Account switcher and last-account cookie still work
- [ ] `pnpm test utils/fetch-email-accounts.test.ts providers/EmailAccountProvider.test.tsx`

## Summary by CodeRabbit

- **New Features**
  - Added smoother, staggered reveal animations to landing-page hero content, video, brand logos, and word-by-word text.
  - Added unified rise-and-blur animation effects for landing-page content.

- **Bug Fixes**
  - Improved email account loading by preventing duplicate simultaneous requests.
  - Footer links no longer prefetch pages automatically, reducing unnecessary network activity.

- **Performance**
  - Adjusted supported image sizes to better match typical display widths.


elie222:
Addressed the CodeRabbit note on the self-hosted footer: those links now also set `prefetch={false}`, matching the main footer.

Skipped the docstring-coverage warning — this repo does not add docstrings for straightforward UI/helper changes.

elie222:
https://github.com/elie222/inbox-zero/pull/3773#pullrequestreview-5241011203 (@coderabbitai)

Declining the `animation-fill-mode: both` / hidden-during-delay suggestion. That hidden stagger is the original Framer `BlurFade` word-reveal; we already tried the “keep opacity at 1” version and it looked wrong. `prefers-reduced-motion` still skips the animation so that path stays visible.

PR #3787, opened 2026-09-17
reply-tracker: Define conversation statuses once, on the rule

elie222:
## What

The thread-status prompt hardcoded its own definitions of `TO_REPLY`, `AWAITING_REPLY`, `FYI` and `ACTIONED`, and only used the rule's `instructions` (the text users see and edit) as "preferences" layered on top when they differed from the default. For every account on defaults the visible definition did nothing.

The hidden definition also contradicted itself. The FYI criteria said "NO questions or requests exist anywhere in the thread"; rule 5 said a fulfilled request "should be FYI (if informational) or ACTIONED (if fully resolved)". That free choice is where production flips between FYI and ACTIONED on identical closed support threads ("ok, that's correct, thx" → FYI; "Yes, thank you for the clarification!" → ACTIONED, both in the same mailbox).

Now:

- **The rule instructions are the single definition of each status.** The defaults in `getRuleConfig` grow one clause each so they carry the distinctions the prompt used to. The prompt keeps only the thread-reading procedure (scan every message, whose promise is it, multi-person threads, clarifying questions, taking ownership, and so on).
- **A fulfilled request is `ACTIONED`. `FYI` is for threads where nothing was ever asked.** Rule 5's FYI alternative is gone.
- **The owner-sent-last gate is unchanged**: FYI is still removed from the prompt and the schema when the user sent the last message.
- **Existing accounts keep getting defaults.** Rules store their instructions at creation, so existing rows hold the old default text. `previousInstructions` records those and `isDefaultRuleInstructions` treats them as default, so those accounts get the new default definition rather than the old one-liner.

## Evals

Two arms, GPT-5.6 Luna via OpenRouter, three repeats each. Baseline is `origin/main` (1e46d6473). Paired exact McNemar.

| set | cases | baseline | variant | discordant (baseline-only / variant-only) | p |
|---|---:|---:|---:|---|---:|
| `__tests__/ai-regression` + `__tests__/eval` thread-status tests | 24 | 23, 24, 24 | 23, 23, 24 | one pre-existing failure, see below | |
| Synthetic thread-status set (private evals repo, dev) | 24 | 24, 24, 24 | 24, 24, 24 | 0 / 0 | 1.00 |
| Real threads, conversation statuses, dev split (majority of 3) | 58 | 43 (74.1%) | 44 (75.9%) | 8 / 9 pooled | 1.00 |
| Real threads, conversation statuses, held-out split (majority of 3) | 29 | 19 (65.5%) | 18 (62.1%) | 7 / 5 pooled | 0.77 |

The real threads are 222 production threads from one mailbox, labelled by the shipping rules engine and corrected by adjudication; conversation-status cases are the 87 whose label is one of the four statuses. Scored under the decided policy (the five closed-support threads production called FYI count as ACTIONED), dev is 46/58 on both arms.

So: no measurable accuracy change either way, which is what the change was meant to do. The dev set cannot resolve anything under about 8 points, and run-to-run spread is 3 to 5 cases. The variant is slightly more stable (cases whose answer changes between identical runs: 7 → 5 on dev, 5 → 4 on held-out). The consistent flips are:

- Gains: `awaiting-reply-16` (owner asked a question and sent last; baseline said TO_REPLY three times), `awaiting-reply-18` (other party said "they'll flag it for review"; baseline said ACTIONED), `actioned-10` (baseline gave three different answers in three runs).
- Losses: `awaiting-reply-20` (last message is an auto-reply saying the address is unmonitored), `awaiting-reply-2` (other party said "I'll try to review later today"; variant says ACTIONED), `marketing-28` (an intro where the owner was moved to BCC; variant says FYI, which is defensible).

The held-out split had never been scored before this and has now been seen once by both arms, so it cannot referee a further prompt iteration.

**Pre-existing test failure.** `identifies FYI when receiving instructions after offering help` expected FYI or TO_REPLY. Luna returns ACTIONED on the old prompt too (baseline failed it 1 of 3 runs, variant 2 of 2 before the test was changed). Under the policy above ACTIONED is the intended answer: the user's offer was answered and nothing is pending. The test now accepts ACTIONED or TO_REPLY and passes.

## Tests

- `utils/ai/reply/determine-thread-status.test.ts` used to test a local reimplementation of the customised-rules filter. It now tests the exported `getConversationStatusDefinitions`: defaults, previous defaults, user edits, disabled rules, empty instructions, and the FYI-omitted subset.
- `pnpm test` on `utils/rule`, `utils/reply-tracker`, `__tests__/eval/prompt-security-core-paths`: passing.

<a href="https://cubic.dev/pr/elie222/inbox-zero/pull/3787?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Improvements**
  - Improved automatic conversation status detection for replies and follow-ups.
  - Threads where a user’s offer of help has been answered are now more accurately marked as actioned.
  - Custom status instructions are better respected, while blank, disabled, or unchanged instructions continue using the appropriate defaults.
  - Status suggestions now adapt to the conversation context, including when the user sent the latest email.
  - Irrelevant status options are hidden when they do not apply.


elie222:
Addressed in 2013e4b: rule 11 now defers to the user's FYI text when FYI is customised, and only states the default boundary when FYI is on the default definition. The prompt builder is exported and unit-tested for the default, custom and owner-sent-last cases.

